PR #647, opened 2015-01-02
Fix crash when mailing a URL in the original story view.

mihaip:
[OriginalStoryViewController doOpenActionSheet] calls showSendTo with a nil
text parameter, but [NBActivityItemProvider activityViewController:
itemForActivityType:] returns a dictionary with the text as a value, and
dictionary literals can't use nil values. Use an empty string instead.

Appears to be a regression from 1002403e81087e16695589e7ad1b322856252d40.


samuelclay:
Looks like this may be what's causing an issue for some users. Thanks @mihaip! Will submit to Apple by EOW.

Issue #637, opened 2014-12-23
Android: Reading Progress Broken in Folders


dosiecki:
The progress circle in Reading seems to not be tracking unread counts correctly while reading folders.  Individual feeds seem okay, though.

dosiecki:
All fixed up in #662.  Feel free to re-open if I missed an edge case.

PR #654, opened 2015-01-21
Android: Bugfixes and Speed

dosiecki:
Big round of bugfixes, speed tweaks, and cruft cleanup.  The old legacy DB is finally gone and we are 100% on the raw sqlite driver. Patched a  bug that will free up GB worth of data on some devices.  Made DB auto-rebuilds more reliable and less impactful to the UI.

Lots of changes here, would recommend at least a few days in Beta.


dosiecki:
Oh, and as with all patches, don't forget to bump your version strings!

samuelclay:
Will do. This is working great. Fast and accurate. Now we just gotta fix #637. Should be an easy fix, as you just need to use the folder's total unread count, not the count of stories loaded.

samuelclay:
Should I be disabling logging somehow?

dosiecki:
> Should I be disabling logging somehow?

Nope, already done. (you can always check the top few constants in AppConstants)

Issue #653, opened 2015-01-20
Android: Use volume buttons as story next/previous


samuelclay:
From: https://getsatisfaction.com/newsblur/topics/android-next-prev-button

manderson23:
I'll pick this one up.

I assume we want to hook it into the same actions as pressing the next/prev overlays?

dosiecki:
I don't really understand the use case, so I can't say if the buttons should imitate a left/right swipe or if they should trigger the nextunread/previous buttons.  Also not sure which of up/down maps to which action depending on the orientation of the phone.  I see about 16 different possible algorithms we might choose from.

If we do implement this, make it an option that is off by default.  Default users will be unhappy if we steal away the ability to adjust volume while in the app.

manderson23:
Yeah it isn't something I would use but if @samuelclay wants it. Maybe he has some thoughts.

I can't quite remember but I think [GrazeRSS](https://play.google.com/store/apps/details?id=com.grazerss) supports this so I could maybe see how they handle things.

samuelclay:
Up = next, down = previous. Off by default.

rrehbeindoi:
It was a feature that Google Reader's app had.  It seemed natural to grip the phone and tap the physical buttons on the side, rather than 2nd hand swiping the screen.  Before Google killed it, I think it had 3 options.  Off,  up=next down=prev, down=next up=prev as the physical volume button being on the left or right side of the device changes the feel of how it maps.

samuelclay:
Will launch next version.

manderson23:
From https://getsatisfaction.com/newsblur/topics/android-volume-keys-for-next-previous

"Would it be possible to have the keys act as left/right swipes?"

PR #656, opened 2015-01-28
Android: Volume key navigation

manderson23:
Implements #653 

Since I was adding a preference I went for the three options: off, up = next or down = next.


samuelclay:
Sweet, thanks Mark!

PR #658, opened 2015-02-04
Android: Global Shared Stories

manderson23:
I'd be grateful if @dosiecki could give this a review to see that it fits into the app as expected.

The global shared stories seem to be correct though when I was comparing with the web I noticed that the timestamps on the web were out of order so didn't exactly match what I was seeing in the app. 

I'd also appreciate some wider testing particularly to ensure that All Shared Stories and individual shared stories feeds work as expected.


clients/android/NewsBlur/src/com/newsblur/activity/GlobalSharedStoriesItemsList.java
```diff
@@ -0,0 +1,93 @@
...
+import com.newsblur.util.DefaultFeedView;
+import com.newsblur.util.FeedSet;
+import com.newsblur.util.PrefConstants;
+import com.newsblur.util.PrefsUtils;
+import com.newsblur.util.ReadFilter;
+import com.newsblur.util.StoryOrder;
+
+public class GlobalSharedStoriesItemsList extends ItemsList {
+
+	private ContentResolver resolver;
```
dosiecki: Whoops, this member was cruft in the SavedStories activity.  No need for it, and I have cleaned it up from the original.

dosiecki:
Minus one mistake that was mine, looks great!

PR #667, opened 2015-03-03
Android: Bugfixes and UI Tweaks

dosiecki:
Fixing a few little bugs found in this release.  Also tweaking the UI for story list loading per @samuelclay's request, and added the requested use of feed icons in the actionbar.

As with all releases, I'd suggest this get a week in Beta before going to prod.


samuelclay:
Thanks Daniel, I'll merge this in next week as I'm on vacation in NYC all week.

dosiecki:
Having run this for a week now, on one of my devices I am getting slight memory problems.  Let's definitely test this for a week to see if any of y'all can reproduce, just to eliminate any of these changes as a root cause. (unless I find it before then)

samuelclay:
Ok, releasing as 4.2.2b1 (89).

manderson23:
When you navigate away from the main view should the actionbar icon always update for consistency? e.g. a folder, saved stories etc. instead of only changing for reading an individual feed?

dosiecki:
For starters, I just went after the specific feature request to use feed icons. If we can track down high-q versions of the folder/group icons, they'll be trivial to add.

samuelclay:
Those icons are here: https://github.com/samuelclay/NewsBlur/blob/master/media/img/icons/circular/g_icn_folder_rss.png.

You can see the list of all iOS assets here: https://github.com/samuelclay/NewsBlur/blob/master/clients/ios/Resources/g_icn_folder_rss%402x.png.

I can easily make larger icons, just let me know the size you need. I have vector originals.

dosiecki:
Yeah, that size would look a bit jagged on bigger tablets.  Incredibly, the latest devices will need them at 128x128, with the next rumoured bump going to 192x192.

samuelclay:
These feed icon are way too big. Can you constrain them to 32x32? That way we can reuse the existing icons.

dosiecki:
The system determines the rendered size of the actionbar icon. (note that it is a Drawable, not a View).  There is a trick that can be used to make them an absolute size in pixels (not dp), but that has the nasty effect of them being big on old devices and incredibly tiny on new devices.

In order to gain full control over sizing, we would have to locate them outside of the actionbar or write our own actionbar widget and associated functionalities (like non-conext menuing).

samuelclay:
Can we mimic the size from the feed list? It does need to be much smaller than it is, though. It's the height of the bar.

dosiecki:
Not using a standard actionbar.  The feed list is our widget, so we can define size in terms of dp, but the actionbar uses a set of proportional presets or raw px sizing.  The upper-left icon was intended to have a fixed look and feel across apps, so the raw view isn't exposed.

It is possible to add a more customised view in between the icon and the menu dots, though it is some work to get to look tidy on all devices.

dosiecki:
Let me know if you would like to see a prototype of either approach.  Putting an icon between the app icon and menu should be an hour or so, and replacing the app icon with a custom view should be 2-3 hours.  Full testing and implementation in either case is more of an unknown because the side effects are extraordinarily dependent on screen size, screen density, and platform version. (in the latter case, we actually have to implement our own handling of up- and back- navigation)

samuelclay:
Yes, I would like to see this. Use whichever technique you think will work better.

PR #673, opened 2015-03-16
Android: Custom ActionBar Styling (and Bugfixes)

dosiecki:
Replaced the system ActionBar UI with one of our own so we can control the size of the icon in it.  Also added styling for the remaining reading types.

Also inclues a few little bugfixes found over the last week.

Mucking with the ActionBar can be tricky, so I would want to test this out on most of the Android versions we support, 11-22.  I've tested on 19 and 21 and will get 22 tonight.

### DIFF
--- a/clients/android/NewsBlur/res/layout/actionbar_custom_icon.xml
+++ b/clients/android/NewsBlur/res/layout/actionbar_custom_icon.xml
@@ -5,8 +5,8 @@
 
     <ImageView
         android:id="@+id/actionbar_icon"
-        android:layout_width="25dp"
-        android:layout_height="25dp"
+        android:layout_width="24dp"
+        android:layout_height="24dp"
         android:layout_alignParentLeft="true"
         android:layout_centerVertical="true"
         android:layout_marginLeft="10dp"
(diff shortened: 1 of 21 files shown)


dosiecki:
Have also tested on 16, 17, and 18.  I would really like to test on 11-13 if we are going to continue to support the half dozen or so users on those versions, but my old 3.x device kicked the bucket last year.

samuelclay:
Did we lose the back arrow on the feed detail view?

dosiecki:
Yep, the home-as-up affordance was part of the default ActionBar.  I kept the old behaviour by implementing home-as-back on the custom icon, even though it isn't technically a home icon any more.  The latest Android design guidelines eliminate the use of custom icons there and just replace them all with an arrow, but obviously, the whole point of these changes is to have a custom icon.

manderson23:
I noticed that the folder icon on the action bar is different from the folder icon used on the main view (main view doesn't have the RSS icon within the folder). Is that intentional?

dosiecki:
Folder icons change depending on whether they are open or closed.  I just picked one of the two, since there is no such context in the latter activities.

manderson23:
Ah OK. Don't think I've ever noticed that!

PR #679, opened 2015-03-31
Android: Folder Hierarchies and Bugfixes

dosiecki:
Big rework of folders to finally make the app understand that they have a tree-like structure.  Lots of associated cleanup as a result.  Primarily allowed us to fix the way stories show in nested folders, but also enabled a few other little tweaks.

Also includes a few other bugfixes and speed improvements.


dosiecki:
This now also has a prospective fix for https://getsatisfaction.com/newsblur/topics/android-feeds-wont-update-when-first-opened

samuelclay:
Sorry for sitting on this for a week now. I've been distracted with a non-newsblur gig that last a week and now a non-newsblur project that has been taking most of my time. Releasing as beta momentarily.

samuelclay:
Bumping version code to 90 and then got this screen on load, after "On its way" finished:

screenshot_2015-04-06-14-26-14

Taping a folder collapse button crashed the app, but the log wasn't recording for some reason. And now I can't get out of this state.

dosiecki:
Did the version name also get bumped, or just the version code?

dosiecki:
Oh, wait, think I found it.  Handling for the social folder needs a tweak.

PR #680, opened 2015-04-01
Android: Okhttp

manderson23:
I've been using this without any issues for the past week.

I think there is scope for cleaning up the interaction between APIManager and APIResponse but that could be better handled as a task to look at using the support in Okhttp for parsing a response directly with GSON.


dosiecki:
> I think there is scope for cleaning up the interaction between APIManager and APIResponse 

I have intentionally been putting that off until we decide on whether to switch to an alternative DB solution.  Many of them have integration with JSON parsers that might be of benefit.  (for instance, I think there is a recipe out there for direct GSON->Realm streaming)

clients/android/NewsBlur/src/com/newsblur/util/ImageCache.java
```diff
@@ -52,21 +51,7 @@ public void cacheImage(String url) {
...
-                    // redir will only be set if the original load encountered an un-followable
-                    // redirect. Try once and only once to follow it manually.
-                    u = new URL(u, result.redirUrl);
-                    result = NetworkUtils.loadURL(u, outStream);
-                }
-                size = result.bytesRead;
-            } finally {
-                outStream.close();
-            }
+            long size = NetworkUtils.loadURL(new URL(url), f);
```
dosiecki: Were you able to confirm that okhttp correctly handles redirects both from http->https and https->http?

manderson23: I didn't see any missing images in my testing.

One thing I noticed was that downloading images is really slow (I don't normally have it turned on). Both the old code and the new code. Maybe it is slow storage on my N4?

dosiecki:
> downloading images is really slow (I don't normally have it turned on). Both the old code and the new code. Maybe it is slow storage on my N4?

Since it relies upon both network and IO, yeah, it can be pretty slow in many circumstances.  Plus, it is done at the absolute-lowest priority possible so it doesn't degrade UI-blocking IO operations and cause lag.  Also really depends on the image servers being hit.  Some like to throttle repeated calls, others are smooth sailing.

(weirdly, it is much faster on my N4 than my N7. go figure)

samuelclay:
I just deployed v4.2.3. Is this coming up next?

dosiecki:
Just got this, and it appears to break proguarded builds - did you have to tweak any local environment settings to get it going?

samuelclay:
I did, I had to add the libs. But that was it.

manderson23:
@dosiecki I see that Okhttp 2.4 is out. Do you want a PR to move to that version?

dosiecki:
I generally push out updated dependencies only after a good few weeks' testing on my own devices, unless they fix a critical bug we're actively addressing.  Especially with something like the HTTP lib, I would want to have used it in a variety of (awful) network conditions before sending it out to Beta.

manderson23:
OK. Just wondered since we were moving to it for the next release whether we might as well move the latest. I'll leave it for now.

PR #681, opened 2015-04-06
Android: Fix crash in #679.

dosiecki:
Good catch. Ready for more testing.


samuelclay:
I've gone home for the day with a headache but I'll get this launched tomorrow.

samuelclay:
Deployed as v4.2.3b4.

dosiecki:
Three days and a little over 300 installs in beta and no crashes reported.

If you and @manderson23 have tested to satisfaction, I'd suggest we consider pushing this up to Prod, as it has the one nasty bugfix in it.

samuelclay:
I'm going to wait until Monday to do so, but will then. I'm on vacation in a week, but that should give us 5 days to catch anything should something arise.

manderson23:
Everything fine for me.

PR #683, opened 2015-04-17
Fix :username typo in api.yml

samford:
I was looking through the API and I noticed that `/social/statistics/:user_id/:name` seems to stand out in regard to the similar endpoints around it.  The other endpoints follow a `/social/something/:user_id/:username` format and they all have a Tip that says "The `username` url parameter is optional. It's just recommended if you have it."

This led me to believe that maybe the `:name` url parameter was maybe just a typo, so here's a pull request that fixes it in case I'm correct in my thinking.

### DIFF
--- a/templates/static/api.yml
+++ b/templates/static/api.yml
@@ -919,7 +919,7 @@
         - >
           The <code>username</code> url parameter is optional. It's just recommended 
           if you have it.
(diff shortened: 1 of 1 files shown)


samuelclay:
Nice catch!

PR #690, opened 2015-05-05
Fix release builds that use okhttp. (#680)

dosiecki:
Added the necessary config so release builds will work with okhttp.


manderson23:
Sorry wasn't aware of this. I'm building from within Android Studio.

PR #691, opened 2015-05-06
Android: Beta Bugfix

dosiecki:
Fixing a crash reported from the current beta.


samuelclay:
Awesome, I'll release this along with a fix for: https://getsatisfaction.com/newsblur/topics/android-long-press-on-item-with-parenthesis-crashes-4-3-0beta1

dosiecki:
Looking at all the crash logs, I believe that the parens were a red herring and this fix addressed the underlying problem.

PR #694, opened 2015-06-05
Adding Opera Mini as an option for default browser.

AlphaCluster:
While Safari and Chrome are decent enough browsers sometimes you want to use less bandwidth. So I added a way to set opera mini as the default browser as well.


samuelclay:
Nicely done.

PR #696, opened 2015-06-05
Android: Fix AS builds?

dosiecki:
Give this a try to see if it fixes AndroidStudio release builds.


dosiecki:
Scrap that.  Installing AS real quick to dig on this.

PR #697, opened 2015-06-05
Android: Fix AS Release Builds

dosiecki:
Turning off proguard entirely, since it has broken 2 of our 3 last releases.  (since we ditched the old ABS libs, it isn't actually doing that much for our code anyhow!)

*Be sure to check your AS/Eclipse project config to ensure it won't override project config and run proguard anyhow.*


samuelclay:
No dice. ProGuard is turned off in project settings.

dosiecki:
Do you see in your build logs that proguard is still being run?  The release APK should be more like 885kB if it is correctly disabled.

samuelclay:
899kb, and I don't know if I ever ran ProGuard, as the other APKs are just a few KB smaller.

samuelclay:
And I left Proguard unchecked during the APK signing process.

dosiecki:
Does your AS install know about all the jars in the libs/ directory?

I am a bit stumped as a non-AS-user.  Perhaps @manderson23 can pull from HEAD and advise?

dosiecki:
Just stabbing at things, here: try an AS `Build > Clean Project`, delete all apks that are hanging around, double check settings, and `Build > Rebuild Project`?

samuelclay:
I've been rebuilding the entire time.

samuelclay:
And there are 5 jars in libs, all properly included. Everything does compile (albeit with 23 warnings).

dosiecki:
If you grep for `minify` or `proguard` on your (clean) project directory, do any config files have suspicious hits?

samuelclay:
Already greed for program, nothing. Same for minify.

dosiecki:
Still stumped.  I totally `rm -rf`'d my AS install and my NB git clone and started from scratch.   Re-downloaded AS 1.2.1.1, cloned the full NB repo from HEAD, imported the project, created a release build with all default options, and the release builds fine and the apk runs fine on 4.4 and 5.0 devices.  The only thing I changed in the setup wizard is I told it *not* to muck with the libs, but to just leave them as-is.

Might need to upgrade or re-install?

manderson23:
You need to configure the IDE for Butterknife http://jakewharton.github.io/butterknife/ide-idea.html .

On Mac ```Android Studio -> Preferences```, select ```Build, Execution, Deployment > Compiler > Annotation Processors``` from the tree. Check ```Enabled Annotation Processing```.

manderson23:
Apart from adding the butterknife jar to the project libraries that is all I needed to do.

samuelclay:
Enabling annotation processing did the trick. Mark, is there any way we can spit out a debug message if that isn't turned on?

manderson23:
I've no idea. The build initially failed for me so I just had a look through the preferences and project settings. When I saw that option I wondered if Butterknife required any annotation processing and just tried it :-).

dosiecki:
FWIW, the newest AS enabled it by default on import. It might be a best practice to just cleanly re-import any time libs are added. (I essentially nuke and recreate my env on every build, too)

PR #699, opened 2015-06-06
Add login extension support

AlphaCluster:
I set up the login page to implement the login extension using the 1Password extension.


samuelclay:
Holy hell, this is beautiful. Thanks so much for the patch! It'll be included in the next beta which is coming up in a week or two (has a bunch of nice additions).

Issue #700, opened 2015-06-07
Android: Latest System WebView Changes Reading UX


dosiecki:
On devices running Android version 5.0 or greater, the system WebView component is now updated more regularly, out-of-sync with overall OS updates.  The latest update pushed out to users (43.0.2357.121) greatly changed the behaviour of HTML links within an app.

The old, desired behaviour of opening links in the user's chosen browser has been replaced by opening the link within the same WebView View within the app.  This has caused countless bugs:

* Pages not conditioned (with our custom CSS) to be viewed in-app tend not to render entirely on-screen
* Scrolling of content can be badly broken
* Back-button functionality is broken - if a link is clicked, hitting back exits the Reading activity rather than going back to the page that contained the link
* In-app browser state is not shared with the user's browser, so logins don't work, nor does history get populated
* Much JavaScript tends to be broken in a constrained app View

I've had several reports from people saying this makes the app "unusable", so pursuing a fix for this ASAP.

samuelclay:
Wait, which reports? Are you just getting direct emails or does this show up somewhere that's not GS? Did I miss it somehow?

dosiecki:
Just verbal reports. (I try to convert as many people to NewsBlur as possible!)

Fix is done in 245d0d4.  I am tempted to say this should go into the next prod push, as it is quite an awful experience once your 5.0+ devices get that update.

dosiecki:
Fixed in #702.

PR #702, opened 2015-06-08
Android: Ensure links in stories open in browser on all platforms. (#700)

dosiecki:
Fixing the really broken link-clicking experience that is slowing spreading as users get the related system update.  Should make the update seamless if we get it out with this release.


samuelclay:
To be clear, you want this to go out with the beta build, right? I'll release this with the beta right now, but we don't need to prioritize this before what's in the beta, correct?

dosiecki:
Yeah, with beta, not necessarily before.

samuelclay:
Out as 4.3.0b4

PR #703, opened 2015-06-12
Android: Beta Bugs

dosiecki:
Fixing the two crashes that have been reported in beta and also the quick fullscreen upgrade.  

After a day or two of soaking this shouild probably go to prod, since #700 is spreading rapidly.


samuelclay:
Beta out as v4.3.0b5.

Issue #693, opened 2015-05-21
Android: Interactions and Activities view


samuelclay:
Just like iOS and web.

![screen shot 2015-05-21 at may 21 10 56 02 am](https://cloud.githubusercontent.com/assets/44229/7755310/fd80ae5e-ffa7-11e4-8faa-22fd118ffd52.png)
![screen shot 2015-05-21 at may 21 10 58 04 am](https://cloud.githubusercontent.com/assets/44229/7755356/477f1716-ffa8-11e4-84aa-fb5833079cfb.png)

On iOS:

![img_2798 2](https://cloud.githubusercontent.com/assets/44229/7755342/33218ac4-ffa8-11e4-9b16-bbffbd05020c.PNG)
![img_2800 2](https://cloud.githubusercontent.com/assets/44229/7755346/333b7c0e-ffa8-11e4-8722-23c1bac9dc4f.PNG)

Note that I have a couple bugs to fix on iOS. It should strip HTML on interactions, and it should be loading those icons in Activity.

samuelclay:
See github.com/samuelclay/newsblur/tree/master/templates/reader/interactions_module.html and github.com/samuelclay/newsblur/tree/master/templates/reader/activities_module.html for all of the categories and how they should be formatted.

manderson23:
Thanks

manderson23:
I assume on iOS the ">" takes you to the feed/story etc?

samuelclay:
Correct, and not only that, but it scrolls you down to the comment too.

manderson23:
Is your preference that you can click anywhere to navigate like on the web or that there is a ">" like iOS?

samuelclay:
Tap anywhere, since you can't tap on individual pieces.

manderson23:
So in iOS you can't actually tap on the links? The current Android code was attempting to make each part clickable (e.g. username was clickable on comment reply) but it would simplify things if we didn't need to do that.

samuelclay:
Nope, not at all. They're just styled that way, it actually just takes you to the comment.

manderson23:
@dosiecki I'm looking at navigating from saved story activity details to the saved story itself.
Am I right that given the following details
```
{"category": "star",
 "user_id": 40088,
 "title": null,
 "is_new": true,
 "story_feed_id": 76,
 "content": "\u2192 How Well Can You Hear Audio Quality?", 
"with_user_id": 40088, 
"feed_id": 76, 
[...]

PR #704, opened 2015-06-15
Android: Support passing story hash into Reading activity.

manderson23:
Pulling out some code from work on #693 for @dosiecki to make use of.


dosiecki:
Most excellent.  Will merge this as soon as the new Prod release has been out for 24 hours, just so we don't complicate any emergency bugfixes.

manderson23:
I'll just close this since the same change went in as part of #709

PR #709, opened 2015-06-28
Android: Activities and Interactions

manderson23:
So I've tried to keep this as close the the behaviour as seen on the web as a could. As mentioned previously comment navigation isn't supported. But once @dosiecki has merged his other changes I'll look at adding scrolling to the comment header in a further PR.

The only other difference from the web is that it only supports navigating to feeds/users that the app knows about. It doesn't go searching for any user/story like the web does if you don't follow a user/subscribe to a feed etc.

I'm open to any suggestions/feedback you might have.


dosiecki:
Looks pretty good to me!  As soon as this is pulled, I will merge it into my branch (I suspect there will be some interesting conflicts) with the on-load hooks.

samuelclay:
@dosiecki you're waiting on me to merge this, correct?

dosiecki:
Would be good to have by tomorrow, but not a blocker just yet.

Issue #515, opened 2014-04-07
Android: Videos won't open full-screen


samuelclay:
Trying to watch a video from Adafruit.com (title: The globes that will *change the way we make music*) and I can't full screen it. I've also heard of this issue before from other users.

dosiecki:
Took some doing, but I got this working on a majority of devices in 4ac7732a9bc6cc0846bcd9b1992794dca122f66b.  Look out for it in our next feature release.

dosiecki:
Have tested this will all sorts of HTML5 content including videoes from all the usual places (youtube, vimeo, dailymotion, etc) and even with some embedded VR(!) and the all worked great.

PR #711, opened 2015-07-01
Android - FS HTML5 and on-load hook

dosiecki:
* Fix for #515 so that HTML5 videos and widgets can do full-screen stuff without needing to be popped out into a full browser
* For @manderson23: hook to detect when a story and all related content and comments have finished loading so that custom position-dependent scrolling can be performed (see ReadingItemFragment.java:609)


manderson23:
Thanks

PR #716, opened 2015-07-16
Android: Fix #715

### LINKED ISSUE #715: Android: Volume key navigation broken for global shared/read/saved stories
As reported at https://getsatisfaction.com/newsblur/topics/android-volume-button-navigation-doesnt-work-on-global-shared-stories

Confirmed that it isn't just global shared stories.

manderson23:
Fixes issues with volume key navigation on activities where the next/prev overlay isn't supported.


clients/android/NewsBlur/src/com/newsblur/activity/Reading.java
```diff
@@ -771,12 +771,34 @@ private boolean isVolumeKeyNavigationEvent(int keyCode) {
...
+                }
+            }
         }
     }
 
+    /**
+     * Global shared, saved and read stories don't have overlays enabled for
+     * navigating between stories so we need to fall back to swipe left/right behaviour.
+     */
+    protected boolean overlayNavigationSupported() {
```
dosiecki: I'd suggest changing the name of this.  It isn't that they don't support overlay navigation, it is that they don't support unread-searching, since they don't have unreads.

clients/android/NewsBlur/src/com/newsblur/activity/Reading.java
```diff
@@ -771,12 +771,34 @@ private boolean isVolumeKeyNavigationEvent(int keyCode) {
     private void processVolumeKeyNavigationEvent(int keyCode) {
         if ((keyCode == KeyEvent.KEYCODE_VOLUME_DOWN && volumeKeyNavigation == VolumeKeyNavigation.DOWN_NEXT) ||
             (keyCode == KeyEvent.KEYCODE_VOLUME_UP && volumeKeyNavigation == VolumeKeyNavigation.UP_NEXT)) {
-            overlayRight(overlayRight);
+            if (overlayNavigationSupported()) {
+                overlayRight(overlayRight);
+            } else {
+                int nextPosition = pager.getCurrentItem() + 1;
+                if (nextPosition < readingAdapter.getCount()) {
+                    pager.setCurrentItem(nextPosition);
```
dosiecki: We might need to do some additional checking to ensure that the pager isn't null, and that setCurrentItem() could still throw exceptions despite the check - the async nature of cursor updates makes it possible for the number of available pages to change out from under you in the time it takes to run just one line of code.

dosiecki:
Code looks good to me.

What I'm still unclear on is whether this is the expected behaviour or not.  The volume keys are supposed to mimic what the left and right overlays do, right?  Is it a confusing user experience that for some story lists they do that and that for others they simply go left and right?

manderson23:
True but jumping straight out the feed list is wing too.

Is the real problem then with the overlays themselves and the fact that next isn't supported on some lists since it always says done?

dosiecki:
> Is the real problem then with the overlays themselves and the fact that next isn't supported on some lists since it always says done?

That is very much by design; it isn't a "next" button so much as a "find an unread or go back to feed list" button.  What might be weird is that there is an "unread finder" feature on rivers that don't support unreads.

Didn't the volume keys on GReader just mimic left/right swipes, not unread-finding, IIRC?

manderson23:
In #653 @samuelclay wanted the volume key to mimic the overlays. I'm happy to change it to left/right swipe but it is his call.

samuelclay:
Let's go with left/right in that case. Since right acts as next unread when in unread only mode. So if somebody complains, they can just use that mode coupled with an order they prefer.

manderson23:
Updated to always be left/right swipe.

manderson23:
Is there anything blocking this being merged?

samuelclay:
Sorry about that, didn't notice it. Thanks!

Issue #713, opened 2015-07-10
Android: Fatal API Failures May Block Updates


dosiecki:
We've had several reports over the last few days of Android users being unable to get new stories on their devices.  After working with some affected users, it appears that some API action(s) are getting stuck in the action queue, blocking further API activity.

As a general rule, pending API actions (mark read, save, share, etc) are executed ASAP before other API actions or reads.  If these fail for any unexpected reason, they are kept in the queue to prevent data loss.  To ensure correct order-of-operations, they are executed strictly in the order corresponding to the user actions that caused them.  Should any API action fail repeatedly or in a fatal manner, that action can block the API queue and cause the app to stop functioning properly.

I haven't been able to reproduce any of these fatal API calls, so it is difficult to diagnose what is causing them.  If anyone can identify one or more, we can put together a fix.  There appears to have been a spike in these over the last few days, if that leads to any hints.

As a work-around, the action queue can be cleared by logging out/in of the app.

samuelclay:
I don't have this queue on iOS. If anything fails, it either informs the user with a toast (story failed to save/share/mark unread/delete/rename/train), or silently fails (mark read is the only one I can think of).

We can probably toss this queue or execute it out of order. It's not really that important to get the order 100% correct, since most of the time it's read this, save that. 

I'd reconsider the queue and just get rid of it. Retry a call once and then don't bother.

dosiecki:
Wait, then how does offline work in iOS?  I would hate to lose all my reads / saves / shares when I am reading on the subway.

samuelclay:
iOS doesn't actually check whether it's offline or not. If a mark read fails, it gets added to a db table. When you refresh your feeds (whether by pulling to refresh or automatically after re-entering the app after 10 minutes since last refresh action), that table gets cleared and sent to mark_story_hashes_as_read. If that fails, they are added back into the table. (It may just clear them on success, but that's not a big difference).

samuelclay:
Can't share or save while offline on iOS. But saving would work the same way.

dosiecki:
Ah, okay, that is functionally what the Android app does, though it also happens to perform everything in order and stop if anything fails.  I'll make it so that things keep getting executed even if a call fails. (they are aggressively logged, so if we do ever figure it out, we can still diagnose)

Are there no edge cases where order matters?  Like, would it be a terrible failure to try and mark-read a feed after having unsubscribed to it?

samuelclay:
[...]

PR #717, opened 2015-07-17
Android: Bugfixes

dosiecki:
So, this isn't really any full progress on new features (though some groundwork for them is included), but it does contain a few small bugfixes and the big bugfix for #713, which appears to be pretty active in the wild.  I would suggest we cut an Beta ASAP so folks in Prod can get some relief.  This has tested well for me, so I would put it on the fast track if possible.


samuelclay:
Pushed 4.3.2b1.

Issue #184, opened 2013-04-18
Android: Header bar shows user and username


samuelclay:
Show user profile photo at top left instead of generic user icon and newsblur logo. See iOS screenshot.

samuelclay:
I'd like to see this next.

dosiecki:
Not quite clear on this.  Do you want to replace the title of the app with the username of the logged-in user and the app icon with their photo, or do you want to augment the UI by displaying these two in addition?  I can't quite figure out how to shoehorn the iOS look into the Android platform style in this case.

samuelclay:
I want it to look like the iOS app's header, with the user's avatar instead of the NewsBlur logo, their name, and total unread counts below the name.

dosiecki:
Done in #720.

Issue #183, opened 2013-04-18
Android: Interactions/Activities button


samuelclay:
Interactions button in top right. Shows interactions, which is already built (see User Profile), but can be infinitely scrolled.

dosiecki:
Done in #720.

Issue #180, opened 2013-04-18
Android: Empty Focus/Unread mode explainer


samuelclay:
If no feeds have unreads in Focus/unread mode, show focus explainer (see dev.newsblur.com using a new account, just click on Focus). When there are no stories in that mode, show the explainer and world icon.

Don't be afraid to create test accounts. I just ask that you preface them with `test_`.

samuelclay:
Is this still an issue?

dosiecki:
I fixed it so that the empty view now correctly differentiates between loading / no unreads status, but the message is still very generic.  Popping in some unformatted text snippents would be quite easy, though doing formatted stuff will take an hour or two.

samuelclay:
Something formatted would be preferable. See what I've got on iOS and the web. Use the same image.

dosiecki:
Done in #720.

Issue #178, opened 2013-04-18
Android: Redraw the footer to match iOS


samuelclay:
* Add button on left
    * Manage/settings button on right

dosiecki:
Done in #720.

PR #720, opened 2015-08-03
Android: Bugfixes and Feed List UI Update

dosiecki:
Big round of tweaks to get the Android UI in step with the newsblur.com UI, including #178, #180, #183, and #184.  I still need to do some tweaking (especially of the image resources), but wanted to get this out to Beta as early as possible to see how it works for folks.

Also included a few noncritical bugfixes.


samuelclay:
Getting a crash when going into and back out of interactions/activity view.

samuelclay:
Tapping on your profile photo should open your profile page. Works like this on iOS, although not the web.

samuelclay:
By the way, this looks awesome and I'm so excited this is going to be released soon.

samuelclay:
Can we remove that ~10 pixel white border at the top of the story titles list? It's leftover from the loading indicator, but I'd like to collapse that space when it's done.

samuelclay:
That's all I've got. Next we should fix the Add Site dialog to include a folder destination and autocomplete-as-you-type. Thanks!

manderson23:
Does the menu still need to include Add new feed and Profile items since there are now separate buttons?

manderson23:
On the Main view I see "You have no unread stories" but a blurblog shows 8 unreads. Are unreads on blurblogs deliberately  not counted?

dosiecki:
> Getting a crash when going into and back out of interactions/activity view.

Definitely send these along or report if installed via the market.  I fully expect some tweaks as we test changes this big on the ~100 platforms we support.

> Tapping on your profile photo should open your profile page. Works like this on iOS, although not the web.

Will do.

> Can we remove that ~10 pixel white border at the top of the story titles list? It's leftover from the loading indicator, but I'd like to collapse that space when it's done

Will see if I can rearrange some layers and get rid of it.

> Does the menu still need to include Add new feed and Profile items since there are now separate buttons?

There is about a 50-50 split on whether action buttons should have text-only conterparts for accessibility reasons.  I generally leave them in unless menu space is becoming constrained.

> On the Main view I see "You have no unread stories" but a blurblog shows 8 unreads. Are unreads on blurblogs deliberately not counted?

Good catch.  I need to check how social is wired into the explainer.

Issue #723, opened 2015-08-05
Android: Sanity-Check CPU/Battery Usage


dosiecki:
As the Android app has become more full-featured over the last few years, it has definitely become something of a resource hog when doing heavy reading sessions.  Additionally, it looks like the app very aggressively held CPU resources during conectivity issues earlier this week for some user configurations.

This task is to do a basic sanity check of how we behave in these circumstances, since being a battery-killer is pretty much a cardinal sin on mobile.

samuelclay:
Awesome, this would be great work. Also, I feel like any work spent on eating less battery shows up in perf improvements.

For the record, NewsBlur on iOS eats battery at a similar rate as active web browsing and twitter do. Nothing out of the ordinary like camera or maps/gps.

dosiecki:
After a few weeks of tesing on 4.4.4, 5.0.1, and 5.1.1, this looks to have been a great improvement.  NewsBlur doesn't even show up on the top battery/memory users list on some of my devices anymore.  We'll obviously keep working this down whenever reasonable, but I am calling this done for the sake of the existing complaints.

samuelclay:
Excellent. So where are we on the latest beta? I remember I chose not to submit the latest beta to the play store as it had some showstopping bugs for me.

dosiecki:
I've got a PR incoming that fixes those and the jump-to-activity bugs coming out today.

PR #737, opened 2015-09-01
Android: Efficiency and Bugfixes

dosiecki:
This patch primarily goes after the issues identified in #723.  CPU usage, battery usage, memory churn, and storage usage have all been reduced somewhat.  In cases when the API is unreachable, background CPU and battery usage have been even further reduced.  More importantly, the included design changes will make it much easier in the future to keep reducing CPU and memory overhead.

Also cleaned up a good deal of legacy code and resources and fixed some lint.

Finally, squashed a few spare issues that were identified in the last few weeks.

I anticipate a few more tweaks before we can finally push all this new stuff all the way to Prod, but I think this is a great checkpoint to send to Beta.

There may be a few minor visual changes on some devices as a result of the optimisations.  If any of them are problematic enough to be of worry, feel free to open a new issue and I will see what work-arounds I can find.


manderson23:
Could the UI be getting stuck showing "Catching up reading actions"? It seems to be spending a lot more time showing that overlay (2 mins+). Even sometimes after navigating into a feed and straight back out.

dosiecki:
Only if you are experiencing network issues or have a stuck request. Logs
should be informative.

manderson23:
I also seem to be getting error messages showing in toasts regularly when marking older/newer as read. Though the actions seem to be successful.

dosiecki:
Logs?

manderson23:
I'll try and catch it.

dosiecki:
One thing I fixed was API->user messaging.  It used to be that all API calls failed silently, even if the errors should have been pushed to the user. (the classic example is trying to unmark a story older than 30 days)  There in one known edge case where the API returns an error message that is not meant to be shown, but there could certain be others.  The logs will explain exactly what fields and values triggered the condition.

samuelclay:
Which call is that? I'd be happy to change it to what we do show the user. Alternatively, feel free to submit a PR in the python code if it's a text change.

dosiecki:
I had recalled that `/reader/delete_feed` returned an error message even on success that was something we had to filter, but maybe that did eventually get fixed?

Basically, if an API response has a `message` (scalar) or `errors` (array) field populated, that message will be displayed to the user.

manderson23:
I think delete feed got fixed at some point.

PR #744, opened 2015-09-06
Added a workaround for Linux webkit browsers

zerodogg:
This is the issue referred to from https://getsatisfaction.com/newsblur/topics/does-not-open-links-in-new-tabs-in-various-webkit-browsers

Essentially, WebKit-based browsers on Linux can't handle the way NewsBlur attempts to open tabs in the background. This workaround detects those browsers and disables backgrounding for them.


samuelclay:
Looks good to me. Deploying now...

PR #746, opened 2015-09-07
Removed an unused variable that broke other browsers

zerodogg:
I screwed up the WebKit fix for Chrome/Safari. Here's a fix. Sorry about that.


samuelclay:
How did I not notice `trueSafari`? Anyway, good thing you took this out.

zerodogg:
I didn't either. It was part of my first implementation, but simplified it and forgot to remove it. Sorry about that.

PR #754, opened 2015-09-21
Android: Beta Candidate

dosiecki:
I think we are finally at a point where the big new features and UI have been internally tested enough to cut a Beta for final pre-release testing.  This patch fixes pretty much all the known bugs I had on my list.

Also, this might be a good time to flip to the new Beta mode GOOG implemented that relaxes the GooglePlus requirement to use the Beta version.


dosiecki:
(usual reminder to bump version strings before testing)

samuelclay:
Looking pretty good. Here's the feedback before we go to beta:

* First thing I saw was an empty feed list, no indication of progress, and 0 counts for unreads. It looks a few seconds until it showed any activity. Can we hide the user icon, unread counts, and feed list until we have something to show? On iOS it's just blank with a giant spinner saying loading. We can just ensure that we're showing the feed loading bar at the bottom.
* Why is the first thing we do "Cleaning Up"? Can we move that to the end? I find that it takes a while and loading of feeds is really slow while that's happening. I'd like to have the first experience be a fast one. 
* When loading stories, the story loading bar should be the bottom-most cell, not at the top. Only when there are no stories yet is it at the top. I just swap out the bottom-mode cell in the table view and use the loading bar. Ideally, you'd only see it initially and it would always load stories below your scroll position without you ever seeing it.

Now, I've noticed my Android tablet is slowing down. Is this normal? NewsBlur feels slower every time I use it, and I have a feeling my Nexus 7 tablet is showing its age. But I also assume it matches the hardware other users have. So is a reinstall of the OS the best way to alleviate this?

dosiecki:
> First thing I saw was an empty feed list, no indication of progress, and 0 counts for unreads. It looks a few seconds until it showed any activity. Can we hide the user icon, unread counts, and feed list until we have something to show? On iOS it's just blank with a giant spinner saying loading. We can just ensure that we're showing the feed loading bar at the bottom.

On any reasonably modern device, the lag before the progress indicator fires up should be less than 500ms.  If it is any longer than that, you might actually have a failing storage chip.  The code puts up the sync indicator and message as soon as it possibly can given the speed of the system.  The only way we could make it faster would be to add a full screen before the main one where we do something like a power-on-smoke-test, but that would, ironically, slow down loading for most users.

> Why is the first thing we do "Cleaning Up"? Can we move that to the end? I find that it takes a while and loading of feeds is really slow while that's happening. I'd like to have the first experience be a fast one. 

Did it say "cleaning up" or "tidying up"?  The latter is the app upgrading because you just installed a new version and cannot be skipped.  The former is a low-priority background task and means that all other high-priority sync tasks are already done and are not being blocked.  Should also note that for first-time installs, there is no upgrade, so it won't happen and first experiences are immediate.

> When loading stories, the story loading bar should be the bottom-most cell, not at the top. Only when there are no stories yet is it at the top. I just swap out the bottom-mode cell in the table view and use the loading bar. Ideally, you'd only see it initially and it would always load stories below your scroll position without you ever seeing it.

The loading indicator moves to the bottom of the story list as soon as sync starts.  If it is staying up top for a tangible amount of time, there is a problem getting to the network (but if you are totally offline, it should disappear immediately. transient network states in Android are awful to detect).

> Now, I've noticed my Android tablet is slowing down. Is this normal? NewsBlur feels slower every time I use it, and I have a feeling my Nexus 7 tablet is showing its age. 

Ah, the Nexus 7 is in a bad place these days.  The 2012 version and the 2013 version have both started exhibiting really bad hardware failures as they age. :(

http://appleinsider.com/articles/13/06/18/googles-nexus-7-tablets-dying-early-possibly-due-to-cheap-memory

http://www.androidpolice.com/2015/04/19/reports-suggest-2013-nexus-7-units-are-failing-at-an-alarming-rate-not-likely-a-result-of-firmware-updates/

Updating software won't help much.  That particular model of tablet is just going bad.  You can even see the decay via install statistics:

clipboard01

samuelclay:
Ok, I'll buy the upcoming Nexus 5 or 6 that comes out next week. As for the beta, it's about to be up as 4.5.0b2. I'm writing the release notes. Fixed activities and interactions, new UI on the feed list to match the web, efficiency and bug fixes. What else in this release?

dosiecki:
* massively reduced CPU and battery usage
* fully functional activities/interactions list
* new feed list UI
* "explainers" for empty lists
* support for more types of alt/title text on images
* will actually tell users if they try to do something impossible (like un-mark a story older than 30 days)

and yeah, tons of bugfixes, especially around offline and what happens when the API can't be reached.

samuelclay:
Pushed to the new open beta. Will tweet tomorrow morning.

dosiecki:
Nice.  Will monitor closely overnight for crashes.

PR #785, opened 2015-10-24
Android: Platform v6.0 Fixes

dosiecki:
Bundle of fixes for devices running Android 6.0.  Also enabling 6.0-specific features for beta testing.

This should go to Beta, but not necessarily to Prod.

*Note: this contains a platform version bump, so if your dev environment overrides the project.properties file, be certain you are now building on API v23!*


samuelclay:
🙌 Pushed to beta channel.

dosiecki:
Excellent, will monitor.  Some of these bugs were fairly annoying to the growing population of users on 6.0, so if anyone reports problems from Lolipop, we should direct them to give the Beta a whirl.  As with any platform version bump, though, I hope we can keep this in Beta for as long as tolerable before going wide.

samuelclay:
Yeah, it took a bit of finagling to get the new platform working. Had to run the SDK manager three separate times, but I think we're now finally running on v23.

samuelclay:
I was on v19 until now, in case that has any effect.

dosiecki:
Nope, we should all have been on v19 up until this change.

If you have the bandwidth for it, sometimes the easiest way to pick up a platform bump is just to obliterate your entire AndroidStudio or SDK install and restart with an install of the newest version and reimport the project. (?!)  Luckily, we generally only need to do this once a year.

Issue #781, opened 2015-10-20
iOS: Press and hold for img alt text broken on iPhone 6s


samuelclay:

Dejal:
I can't reproduce this.  It seems to work:

![img_0185](https://cloud.githubusercontent.com/assets/1318849/10684840/db30b74e-7906-11e5-9e36-797722385e6e.jpg)

acdha:
Maybd it's only a problem on the 6S Plus? On mine, using either the latest release version or the TestFlight 4.9.3, nothing appears to happen when I press on an image – no menu, highlight, etc.

Dejal:
I don't have a 6S Plus to test on, but that screenshot is my 6, which works.  I also tried a 6S and 6S Plus in the iOS Simulator, and it works fine there.  Maybe it's specific to the feed.  Can you give me an example of a feed and story that doesn't work for you?

acdha:
It's a problem on every feed. I notice it most on https://newsblur.com/site/2318003/xkcd.com

Dejal:
A great example, since they often have long alt text.  But it still works for me both on my 6 and the 6s Plus simulator.  Weird.

I'm trying to think if there's some usage reason... are you trying it in the story view, or a browser view?

acdha:
It works in the in-app browser and in-app Safari. It does not work in the Story view and, now that I tried it, neither does the normal URL menu – e.g. long-pressing on a link does nothing, but it does open normally if you just tap it.

Using the same account on the same feed on the old iPhone 4S (it's _so_ small!) which I migrated everything on the 6S+ from, the long touch works normally.

samuelclay:
Chris, are you on iOS 9.1 or 9.0?

acdha:
9.1 now, but it happened with 9.0.1 and 9.0.2 as well.

samuelclay:
Since I don't know how 3d touch works, are you trying both a 3d/force touch as well as a non-3d long press? Is there any difference? What happens when you do one or the other?

acdha:
I've tried all three, and also after disabling 3D Touch entirely in the settings (it is, to be honest, not much more than a gimmick currently).

On my iPhone 4S when I long-press there's a brief flash of a rectangular translucent bubble before the menu displays. On the 6S I either see nothing or, rarely, see that bubble but the menu doesn't display:

[...]

PR #787, opened 2015-10-25
Make 3D Touch work to bring up contextual menus.

nriley:
Also makes delay-based contextual menu trigger work on 3D
Touch-capable devices (#781).


samuelclay:
@nriley, you rock! Thansk for submitting this. We'll get it into the next beta which should be out in the next few days.

Issue #734, opened 2015-08-31
Android: add preference to auto-open first unread story.


samuelclay:
From https://getsatisfaction.com/newsblur/topics/android-app-directly-open-the-first-story-by-opening-a-folder

We have this on the web, so why not have it on Android?

dosiecki:
Should this only work in folders or also in individual feeds?

samuelclay:
Works on both (as well as on blurblogs). Obviously it only finds the first unread if there is one and doesn't kick you out if it doesn't find one.

dosiecki:
So if you tap into a feed with no unreads, does it stay on the story list or does it open the first (already-read) story?

samuelclay:
It stays on the story list. You can double check this behavior on the web.

dosiecki:
Sounds good.  Implementation incoming.

dosiecki:
Done in #795

PR #795, opened 2015-10-30
Android: Beta 2 for Lollipop Compat

dosiecki:
Another big round of fixes to get us running smoothly on the newest version of Android.  Fixes several crashes and some nonsensical behaviour resulting from the OS's naïve attempts to automagically back up all the wrong things.

Also snuck in the impl of #734, which required us to pay off some old technical debt, but that fit nicely with some of the compat fixes.

This should be good to push to Beta so all the existing Beta users get it automatically. (wouldn't go to Prod, though)


samuelclay:
🙌 Uploaded beta 2 to the store.

PR #799, opened 2015-11-02
Docker friendly setup

jrottenberg:
Ideally, we should be able to run the whole site with one command.

The session cookie and the site.site entry syncdb creates got in the way, I'm using docker.newsblur.com to work around but there might be smarter to do.

At least I got it to run locally, after :

```
$ docker exec -it  newsblur_newsblur_1 bash
$ ./manage.py syncdb --all --noinput && ./manage.py loaddata docker/data.json
```

make sure docker.newsblur.com points to 192.168.99.100 

NOTE: need to clean that up.


samuelclay:
Sure, I'll add it, but I can't give official support as I don't use Docker myself. Also, isn't Docker not meant for postgres/mongo/redis/elasticsearch? Just for the app and task servers?

jrottenberg:
Oh thanks that was quick ;-)

Your main app is containerized now, as were postgres, ES, mongo or redis, docker-compose allows us to start all of those with one command, hopefully lowering the effort to get a new developer involved with the project.

I didn't touch the task server as I was not sure how to proceed, and it was working without.  But since you are open to it, I'll give it a shot.

PR #814, opened 2015-11-12
Android: Beta Fixes

dosiecki:
A few fixes and small tweaks for the Beta release, including a particularly irksome crash reported by quite a few of our testers.

### DIFF
--- a/clients/android/NewsBlur/AndroidManifest.xml
+++ b/clients/android/NewsBlur/AndroidManifest.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.newsblur"
-    android:versionCode="108"
-    android:versionName="4.6.0b2" >
+    android:versionCode="109"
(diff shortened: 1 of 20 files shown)


dosiecki:
Also went ahead an added the new folder chooser fragment, since it is testing along nicely!

samuelclay:
Excellent, looks good on both lollipop and marshmallow (I just upgraded my N7). Pushed out as beta3.

dosiecki:
This has looked pretty stable over the last few days, so I think we ought to push it to production to alleviate the crashes being experienced by the ever-growing set of users on Android 6 even though it is a bit of a smaller release.

patch notes:
* full Android 6.0 compatibility
* bugfixes
* auto-open first unread story option
* richer send-to functionality
* choosing folders for feeds

samuelclay:
I'll do that, but I'm going out of town starting Saturday. If we experience issues, we'll have to fix it tomorrow. But I'm OK with that, esp. since this is a patch release. Releasing now.

samuelclay:
Released!

dosiecki:
Excellent.  Will watch closely tonight and tomorrow AM.

Issue #832, opened 2015-11-28
iOS: Don't scroll when going back to a story list


nriley:
NewsBlur on iPhone has a behavior whereby when returning to a story list, it often appears to 'jump' so the previously selected story is in the middle of the screen.  This is disconcerting in a "who moved my cheese" way, as the list is not in the same place as you left it; it also has a side effect whereby the search bar pops up unexpectedly  after a short delay if you're near the beginning of the list.

Both of these problems seem to be solved by one small change.

PR #833, opened 2015-11-28
iOS: Don't scroll when going back to a story list (#832)

nriley:
See #832.


samuelclay:
Very nice. Tested this and it works better than the middle strategy.

PR #835, opened 2015-11-28
Text importer

sv0:
fixed errors that happen when rss content has non-ASCII symbol or encoding is not utf-8.
e.g. cp1251 (windows cyrillic).

### DIFF
--- a/apps/rss_feeds/models.py
+++ b/apps/rss_feeds/models.py
@@ -24,6 +24,7 @@ from django.core.urlresolvers import reverse
 from django.contrib.auth.models import User
 from django.contrib.sites.models import Site
 from django.template.defaultfilters import slugify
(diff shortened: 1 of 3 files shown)


samuelclay:
This is quite the change. How were you able to test it? Do you have NewsBlur running locally?

samuelclay:
This is now launched. Everything's look good, so let me know if this changes a feed and preferably which feed has now been fixed.

sv0:
Yes, I have NewsBlur running locally.
Errors happened while parsing these feeds:
http://24tv.ua/rss/all.xml
http://habrahabr.ru/rss/interesting/
http://www.pravda.com.ua/rss/

samuelclay:
Thanks for this PR! Tweeted: https://twitter.com/newsblur/status/671827327999840256

PR #839, opened 2015-12-02
Unicode fixes

sv0:
a bunch of unicode and utf-8 related fixes.
PEP8 fixes.

### DIFF
--- a/apps/reader/views.py
+++ b/apps/reader/views.py
@@ -5,6 +5,7 @@ import redis
 import requests
 import random
 import zlib
+#import ipdb
 from django.shortcuts import get_object_or_404
 from django.shortcuts import render
 from django.contrib.auth.decorators import login_required
@@ -106,9 +107,7 @@ def dashboard(request, **kwargs):
         del request.session['import_from_google_reader']
     
     if not user.is_active:
-        url = "https://%s%s" % (Site.objects.get_current().domain,
-                                 reverse('stripe-form'))
-        return HttpResponseRedirect(url)
+        return HttpResponseRedirect(reverse('stripe-form'))
 
     logging.user(request, "~FBLoading dashboard")
 
@@ -182,9 +181,7 @@ def signup(request):
             login_user(request, new_user)
             logging.user(new_user, "~FG~SB~BBNEW SIGNUP: ~FW%s" % new_user.email)
             if not new_user.is_active:
-                url = "https://%s%s" % (Site.objects.get_current().domain,
-                                         reverse('stripe-form'))
-                return HttpResponseRedirect(url)
+                return HttpResponseRedirect(reverse('stripe-form'))
     
     return index(request)
         
@@ -963,6 +960,7 @@ def starred_stories_rss_feed(request, user_id, secret_token, tag_slug):
 
 def folder_rss_feed(request, user_id, secret_token, unread_filter, folder_slug):
     domain = Site.objects.get_current().domain
+    scheme = 'https' if request.is_secure() else 'http'
     
     try:
         user = User.objects.get(pk=user_id)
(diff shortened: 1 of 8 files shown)


apps/reader/views.py
```diff
@@ -105,9 +107,7 @@ def dashboard(request, **kwargs):
         del request.session['import_from_google_reader']
     
     if not user.is_active:
-        url = "https://%s%s" % (Site.objects.get_current().domain,
-                                 reverse('stripe-form'))
-        return HttpResponseRedirect(url)
+        return HttpResponseRedirect(reverse('stripe-form'))
```
samuelclay: You'll want to revert this part. This is explicitly redirecting the user to the SSL version, which they may not be on. And we need to not only support http but encourage it as many other websites are also http-only and can't be loaded on an https-only NewsBlur.

apps/reader/views.py
```diff
@@ -5,6 +5,7 @@
 import requests
 import random
 import zlib
+#import ipdb
```
samuelclay: <snip>

apps/reader/views.py
```diff
@@ -181,9 +181,7 @@ def signup(request):
             login_user(request, new_user)
             logging.user(new_user, "~FG~SB~BBNEW SIGNUP: ~FW%s" % new_user.email)
             if not new_user.is_active:
-                url = "https://%s%s" % (Site.objects.get_current().domain,
-                                         reverse('stripe-form'))
-                return HttpResponseRedirect(url)
+                return HttpResponseRedirect(reverse('stripe-form'))
```
samuelclay: ditto

apps/reader/views.py
```diff
@@ -1031,9 +1030,11 @@ def folder_rss_feed(request, user_id, secret_token, unread_filter, folder_slug):
         
     data = {}
     data['title'] = "%s from %s (%s sites)" % (folder_title, user.username, len(feed_ids))
-    data['link'] = "https://%s%s" % (
+    data['link'] = "%s://%s%s" % (
```
samuelclay: So I can see why you're making this change, as the feed link should match the scheme. But how did you come up with this? Were you noticing this on the other end?

media/js/newsblur/models/folders.js
```diff
@@ -331,7 +331,7 @@ NEWSBLUR.Collections.Folders = Backbone.Collection.extend({
         url = url.replace('{folder_title}', Inflector.sluggify(this.options.title));
         console.log(['rss_url', this]);
         
-        return "https://" + NEWSBLUR.URLs.domain + url;
+        return document.location.protocol + "//" + NEWSBLUR.URLs.domain + url;
```
samuelclay: Why change the scheme away from SSL for an RSS feed?

utils/feed_functions.py
```diff
@@ -46,17 +49,9 @@ def run(self):
...
-    try:
-        return unicode(tstr.encode('utf-8', "xmlcharrefreplace"))
-    except UnicodeDecodeError:
-        # it's already UTF8.. sigh
-        try:
-            return unicode(tstr.decode('utf-8').encode('utf-8'))
-        except UnicodeDecodeError:
-            return u''
+    msg = "utf8encode is deprecated. Use django.utils.encoding.smart_unicode instead."
+    warnings.warn(msg, DeprecationWarning)
```
samuelclay: :+1:

samuelclay:
Ok, looks good on the unicode front, but why all the scheme changes?

media/js/newsblur/models/folders.js
sv0: It does not change the scheme.
It just uses the same protocol (https or http) by which website was reached.
e.g. even if you do not use HTTPS to access to the Newsblur you always get HTTPS url  in the 'Folder Settings - FOLDER RSS FEED ADDRESS'
screenshot

samuelclay:
Thanks! This is deployed. Let me know how it works out.

Issue #820, opened 2015-11-19
iOS: improve iPad Pro split view support


Dejal:
The sidebar is too wide in landscape split view on iPad Pro.  Should use the narrow class in this situation.

Dejal:
I've been thinking about this.  It's not an easy problem, as the app currently doesn't use auto-layout (for most views) and size classes.

Possible solutions:

1. Take this opportunity to modernize the main views to auto-layout, size classes, and probably storyboards.  Not quite as bad as it might seem at first, since only the feeds, feed detail, page and story views would need to be changed, not the various popovers.  But those are still fairly major views.

2. Detect Split View use and hide the UISplitViewController and switch to an iPhone-like view hierarchy.  Ideally would need to restore the current view and scroll position.

3. Fake it by resizing the UISplitViewController's left pane to the available width (effectively making the right pane off-screen), and push the right pane onto the left one like on iPhone.

Or maybe some other solution.  Any thoughts on this?

The first solution is no doubt the best, since it'd solve the issue, future-proof the app, and probably reduce code and bugs.  It'd also be the most work, but work that probably needs to be done sometime.

Dejal:
A 4th option would be when in Split View multitasking to simply make the left pane narrower, and perhaps omit the thumbnails, and reduce the margins on the story in the right pane.  That'd be the easiest option, and might be good enough.

samuelclay:
Auto layout should be easy enough to build in, since I already converted the only difficult layout (the story traversal buttons) to auto layout. The other screens are just big table views.

Dejal:
Alright, I'll do option 1, modernizing to auto-layout and size classes.

Dejal:
I added a "5.1" branch for this yesterday.  I'll add subsequent non-fix changes to that branch.  You can merge it into master when you're ready to do the 5.1 release.  You could merge @nriley's changes into the 5.1 branch, too.

The split view support works well; when in split view, it uses the iPhone-style views, then switches back to the normal iPad style when leaving split view.  I'll close this now, but feel free to re-open if you find any issues.

nriley:
Both of the #816 issues I reported were actually issues here; there were some remaining references to a UIPopoverController which was never set.  See pull request above.

nriley:
[...]

PR #840, opened 2015-12-05
Fix regressions in #820

nriley:
After these fixes, my keyboard changes should be safe to merge (will rebase my branch on top of this shortly).


samuelclay:
Oh, I should clarify from earlier today. You should totally continue committing directly to branches like on this ticket. Issue a PR only when going to master.

PR #843, opened 2015-12-06
iOS: Fix long-press story training from feed detail view.

nriley:
The active story wasn't being set so this wasn't working.


samuelclay:
Thanks!

PR #845, opened 2015-12-09
iOS: Use force ratio instead of absolute force.

nriley:
https://getsatisfaction.com/newsblur/topics/swiping-on-images-brings-up-image-menu-sometimes


samuelclay:
Thanks for looking into this!

PR #850, opened 2015-12-21
Fix Mail action in "Send to" from the original story view not including the URL

mihaip:
After 1bbb0fc7e04b90c02305335db00d3ffc7fcf0905 the original URL is no longer included as a separate activity item, so we need to include it in the mail body directly.

Mailing from the story view directly not affected, there we include the URL in the text (done by NewsBlurAppDelegate's `showSendTo:withUrl:authorName:text:title:feedTitle:images:`)

### DIFF
--- a/clients/ios/Classes/NBActivityItemProvider.m
+++ b/clients/ios/Classes/NBActivityItemProvider.m
@@ -46,9 +46,10 @@
 }
 
 -(id)activityViewController:(UIActivityViewController *)activityViewController itemForActivityType:(NSString *)activityType {
-    if ([activityType isEqualToString:UIActivityTypeMail] ||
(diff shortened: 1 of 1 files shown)


samuelclay:
What about a subject? Do you think we should lose the original story's title?

mihaip:
Subject appears to be handled by `subjectForActivityType:`

samuelclay:
I wish that were true, but in my testing it doesn't show up. Each share extension works differently and Mail seems to ignore `subjectForActivityType:` and uses the `title` key from the `itemForActivityType` dict.

mihaip:
I'm still seeing the subject show up on iOS 9.2 when running a build with this commit.

samuelclay:
Ok, it's all set for v5.1, which includes four themes (and a new dark mode), iPad pro support, and numerous smaller changes. Expect it around mid to late January.

PR #852, opened 2015-12-21
Use standard mode in the story detail web view.

mihaip:
It missing a doc type, which was resulting in quirks mode being used. This affects font-size inheritance in tables and other rendering behavior.


samuelclay:
Nice catch. Thanks Mihai!

Issue #745, opened 2015-09-07
Android: Search for Stories


dosiecki:
Someone mentioned this on GetSat, and I totally agree this would be wonderful, even if only for saved stories.

samuelclay:
Should be easy enough to implement. It's built so that all you have to do is pass a `query` parameter to the feed/river request and you get back stories just as if they were from the normal feed. So just keep local state of a query. Clear it out when leaving the feed/folder.

dosiecki:
Working on getting all the edge cases for this right, and discovered a web UI behaviour that I found very surprising and wanted to confirm it is working as expected:  when viewing a list of stories that have been filtered by a search query, doing a mark-all-read marks the entire unfiltered feed/river, not just the filtered subset.  Intended behaviour for the apps, too, or could mark-all-read also take the corresponding query parameter so that only what is shown in the list is marked?

samuelclay:
Ohhh, that is an interesting idea! But I don't think I'm going to support that as I don't see it being used much. It's definitely come up a few times in the support forums, but it hasn't come up for a long while.

Although it wouldn't be too hard to handle on the backend. If people want this, though, they'll have to post on the get satisfaction forum.

dosiecki:
Sounds reasonable.

dosiecki:
Another question: should the `query` parameter work on all rivers?  On this one, I'm not seeing any effect: `https://www.newsblur.com/social/river_stories?page=1&order=newest&read_filter=all&query=foo` (it just returns the same stories with or without the filter.

samuelclay:
Sorry, no search on shared stories. Again, this wouldn't be too hard to build, but I just don't see much justification. And nobody has ever asked for it.

dosiecki:
Righto.  Just making sure I was using the API correctly.  I'll ensure shared stories views don't have the search option.

dosiecki:
In my testing, I am noticing that searches are behaving as case-sensitive. Is this intentional? Can I pass some flag or prefix/postfix some characters to the query string by default to make search insensitive?

dosiecki:
Now I am second-guessing myself.  The first time I did this search, it didn't return results for all capitalisations of "Asus", but now that I am running it again, it is:

https://www.newsblur.com/reader/feed/308521?order=newest&f=308521&query=Asus&page=1&include_hidden=true&read_filter=all

[...]

PR #854, opened 2015-12-26
Android: Search and Bugfixes

dosiecki:
Implements #745.  Also fixes a number of bugs in the current release.  

Good spot for a Beta.  Though most of the nastier bugs being fixed are limited to non-premium accounts, some of them are rather visible.

### DIFF
--- a/clients/android/NewsBlur/AndroidManifest.xml
+++ b/clients/android/NewsBlur/AndroidManifest.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.newsblur"
-    android:versionCode="113"
-    android:versionName="4.6.1" >
+    android:versionCode="114"
+    android:versionName="4.6.2b1" >
 
     <uses-sdk
         android:minSdkVersion="14"
--- a/clients/android/NewsBlur/res/layout/activity_itemslist.xml
+++ b/clients/android/NewsBlur/res/layout/activity_itemslist.xml
@@ -2,13 +2,28 @@
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:orientation="vertical" >
+    android:animateLayoutChanges="true"
+    style="?listBackground" >
+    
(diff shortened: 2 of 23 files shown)


samuelclay:
Published as 4.7.0b1 on the beta channel.

dosiecki:
This version has been perfectly stable, crash-wise.  Provided everyone is happy with the results of their testing and usage, I'd declare this ready to promote to Prod.

samuelclay:
Could we somehow put that search bar at the top of every story list? But offset a bit so that it's hidden? You can scroll up to retrieve it, but it's just hanging out there. And keep the "Search" menu item.

samuelclay:
Oh, I also hit this bug: http://stackoverflow.com/questions/26794862/failure-install-failed-update-incompatible. I got around it by just uninstalling the app and reinstalling, but do you ever run into this?

dosiecki:
did you bump both the version number and version code between builds and
upgrades? this is mandatory from an OS perspective but not enforced by many
IDEs or build environments.

dosiecki:
Fortunately, the Google app store has a sanity check that won't let you
upload a conflicting apk without a bump in versions.

samuelclay:
I did bump both build number and version beforehand. Anyway, how hard would it be to get that search bar visible when scrolling up just a touch. (See the iOS app for a working example)

dosiecki:
> I did bump both build number and version beforehand.

Hmm, did you switch from a release-level to a debug- or test- level build?  Or alternatively, switch which computer you did the signed build from? Trying to overwrite an app with a same-named one with different signing keys will also trigger that error (otherwise just about anyone could steal the data from within an app)

> Anyway, how hard would it be to get that search bar visible when scrolling up just a touch.

I would have to do some playing around.  This isn't a design element I have seen very often outside of about one app I can think of offhand. Swiping down from the head of a list is generally used to trigger a refresh on Android.  If we are looking to make the search function more noticable, it would be quite easy to put the search icon in the ActionBar when there is room, rather than having it buried in the menu.  This would match what about 50% of official Android apps do.

samuelclay:
Ok, I'm game for the search icon. Do we have one already? Let me see... look for g_icn_search_black.png under media/img/icons/circular.

dosiecki:
Will do.

Issue #719, opened 2015-07-26
Adding RSS feed error .


sbilly:
Error accoured, when I adding the feed: http://sbilly.com/feed/

[Jul 26 02:28:11] Traceback (most recent call last):
  File "/root/NewsBlur/utils/feed_fetcher.py", line 582, in process_feed_wrapper
    ret_feed, ret_entries = pfeed.process()
  File "/root/NewsBlur/utils/feed_fetcher.py", line 381, in process
    self.feed.save(update_fields=['feed_title'])
  File "/root/NewsBlur/apps/rss_feeds/models.py", line 222, in save
    super(Feed, self).save(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py", line 546, in save
    force_update=force_update, update_fields=update_fields)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py", line 626, in save_base
    rows = manager.using(using).filter(pk=pk_val)._update(values)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 605, in _update
    return query.get_compiler(self.db).execute_sql(None)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/sql/compiler.py", line 1020, in execute_sql
    cursor = super(SQLUpdateCompiler, self).execute_sql(result_type)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/sql/compiler.py", line 846, in execute_sql
    cursor.execute(sql, params)
  File "/usr/local/lib/python2.7/dist-packages/django/db/backends/util.py", line 41, in execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python2.7/dist-packages/django/db/backends/mysql/base.py", line 120, in execute
    return self.cursor.execute(query, args)
  File "/usr/local/lib/python2.7/dist-packages/MySQLdb/cursors.py", line 207, in execute
    if not self._defer_warnings: self._warning_check()
  File "/usr/local/lib/python2.7/dist-packages/MySQLdb/cursors.py", line 117, in _warning_check
    warn(w[-1], self.Warning, 3)
Warning: Incorrect string value: '\xE4\xB9\x8C\xE6\x89\x98...' for column 'feed_title' at row 1

samuelclay:
You're going to want to use the sitecustomize.py directions to force utf-8 instead of ascii. Details are in the readme.

sbilly:
[...]

PR #855, opened 2016-01-06
Update Dockerfile

sbilly:
Update dockerfile for issue #719.


samuelclay:
Is `reload(sys)` necessary? I've been getting away without it.

sbilly:
It's a walk around solution to solve python environment related issue.

PR #860, opened 2016-01-13
Android: Tweaks to Story Search, Bugfixes

dosiecki:
* added the story search action bar item as per #854 
* re-enabled story prefectch for non-premium accounts after doing some testing and not being able to repro the old bugs that caused it to be turned off in the first place
* fixed some crufty old storage code that will unblock the upcoming work on reading by tag


samuelclay:
Beta?

dosiecki:
Yep, should be a good one to test and then promote.

samuelclay:
Released 4.7.0b2.

manderson23:
Given the get satisfaction thread for mark older/newer has been resurrected I'd be prepared for comments re the search icon moving the mark all read icon left and messing with people's muscle memory. 

I've noticed it though I'm sure I'll get used to it. Though maybe an option to hide it......:-)

dosiecki:
It doesn't overy worry me, since if your device has enough screen space to show the search option, it puhes the mark-all button to the left, so if you muscle-memory hit what used to just be the one option, you don't perform any destructive actions.  The mark-all button, if displaced, moves to what used to be blank space.

Like all Android menu bars, the order is purely defined by priority for being shown on screen.  If we move search to the left, then it is mark-all that gets bumped off for smaller screens, and mark-all seems more generally useful.

PR #865, opened 2016-01-24
iOS: Work around iOS 9 issue where training menu doesn’t appear the first time you select text.

nriley:
This has been bugging me for a while; thought it was a NewsBlur bug.  I can't see how this would disturb anything if/when Apple were to fix the underlying issue.

Source of the workaround and more information at http://stackoverflow.com/questions/32685198/.


samuelclay:
Nice catch. I never noticed this myself because I haven't trained titles since the iOS 9 release.

Issue #866, opened 2016-01-25
Android: Crash when mark folder read selected for Global or All Shared Stories


manderson23:
While testing changes for #856 I came across these crashes.

e.g
```
java.lang.ArrayIndexOutOfBoundsException: length=12; index=-1
   at java.util.ArrayList.get(ArrayList.java:310)
   at com.newsblur.database.FolderListAdapter.getGroup(FolderListAdapter.java:233)
   at com.newsblur.fragment.FolderListFragment.onContextItemSelected(FolderListFragment.java:261)
```
@dosiecki is "mark forlder as read" meant to be supported here? I''m pretty much finished with the confirmation changes but would be happy to add this in.

dosiecki:
Yeah, marking folders via context holds is certainly supported and works fine in my most recent build.  We'll need to look into whether that is a regression or an edge case with your account.

manderson23:
OK. I didn't immediately think from the stack trace it was my changes but I'll pick it up and fix.

dosiecki:
That code does tend to be a bit touchy due to the oversimplified way we handle pseudo-folders via constant offsets.  If the bug looks transient, don't stress it too much, since I am rewriting the way that class indexes groups and children in order to support reading of grouped stars.

Issue #856, opened 2016-01-06
Android: Options for confirming when marking feeds/folders as read


manderson23:
Add a preference to control whether a confirmation is required when marking feeds/folders as read with the following options:

1. Confirm on folders and feeds
2. Confirm only on folders (default)
3. No confirmation when marking as read

PR #870, opened 2016-01-28
Android: Options for Mark All Read

manderson23:
#856 and #866


samuelclay:
Sweet, thanks Mark!

samuelclay:
Looks great. I think this will roll into the 7.2 update. Daniel's working on something, although I'm not sure what. I would like to see saved story tags.

dosiecki:
I was about 4/5 of the way through implementing saved story tag reading when I stopped to work on the hyperuser lag issue.  Will pick up on that as soon as we're out of emergency mode for the latter.

samuelclay:
Excellent, thanks for jumping on that.

PR #871, opened 2016-01-29
Android: Hyperuser Improvements

dosiecki:
A big pass at improving the performance of the app for users who have huge numbers of unreads and a high throughput of new stories (what I have been calling 'hyperusers' as shorthand).  In addition to some basic tuning and tweaking of the DB, I also did some major overhauls to the way we store and query stories to remove any O(n) (where n is number of unreads or new stories for an account) DB operations and overall shorten DB lock times by a factor of 2-5 across the board.  Average users with total unread counts below 10k shouldn't notice much difference, but the change for big users should be like night and day.

That said, the touch surface here was enormous.  I want to get this into y'all's hands so we can have at least 3 sets of eyes using it and looking for regressions and so that there aren't any nasty surprise merges lurking for anyone else touching the code.  I wouldn't be surprised if we'll need to do a few more tweaks and fixes before this is Beta-ready.


samuelclay:
I am one of those users, with 10,000 unread stories (who has time to read more than a few dozen a day *sigh*). And this is like night and day. I'm hoping to push this out as a beta in the next couple hours.

samuelclay:
Would you be OK with shipping a beta today or should I wait until Monday?

dosiecki:
> I am one of those users, with 10,000 unread stories (who has time to read more than a few dozen a day sigh).

Heh, I am the other half - I read almost 1000 stories a day, but the standing count never goes over 1k.  It is definitely the combo of the two that really killed the UX on the old DB arch.

> Would you be OK with shipping a beta today or should I wait until Monday? It looks fine to me, although you're probably right in that there's some bug lurking beneath all of these changes that we have yet to find. But that's what betas are for. I think the massive speed improvement will make up for some bug that isn't so obvious or showstopping.

I just did another hour of testing or so and haven't had any issues yet, so I am increasingly happy. I mostly worry that there might be some "have to uninstall and reinstall" bug lurking, which would be pretty bad, even for folk who are willingly in the Beta.

manderson23:
I've just installed so will try and give it a good workout over the weekend.

manderson23:
If I navigate into global shared stories then into all shared stories or all stories the global shared stories appear in both those views.

Refreshing on the folder list seems to reset it and all stories is as expected.

dosiecki:
> If I navigate into global shared stories then into all shared stories or all stories the global shared stories appear in both those views.

Interestingly, this isn't a regression; I recally tabling the issue quite a long time ago (the selection filter for All Shared Stories is any story with a shared ID and global shareds have those . . .). That said, the new DB arch will finally make this much easier to fix!

manderson23:
Obviously I've never done that before then!

Another one. Open app, all shared stories, back, all stories. Shared stories then appear under all stories. The same issue?

dosiecki:
> Open app, all shared stories, back, all stories. Shared stories then appear under all stories. The same issue?

Similar - shared stories is a subset of all stories, so they certainly are valid stories.  You should be getting other, newer stories from all before you ever see those, though, unless you are offline.  Is your network particularly slow?

dosiecki:
I'll note that across the board, I didn't attempt to fix any issues like this with this release - these are all things that worked this way before this pull.

manderson23:
OK. Again I don't tend to use all stories view much so probably just haven't noticed before.

manderson23:
No problems so far. Although I'm not the target user type for the optimisations it definitely "feels" faster on my Nexus 4 moving into folder/feed lists.

dosiecki:
Have found 2 little bugs to do with read stories incorrectly sneaking into unread-only views after a mode switch. Squashing those today.

Have also had one instance of the DB still going into a deadlock / spinlock, *but* this is not necessarily a regression; it was something I had hoped to have improved for hyperusers.  It is more rare now, but still can happen on very old devices.  (I have only been able to trigger it on a device running DB v3.7.11, which despite being 4 years old, is not unheardof in the wild)  I think chasing bugs for DB engines that old is a bottomless timesink, so I am going to briefly focus on ensuring we didn't regress in the new design and then table the matter.

PR #874, opened 2016-02-04
There is a cert mismatch on popular.global

stevenatwork:
eg. https://popular.global.newsblur.com

### DIFF
--- a/templates/social/social_page.xhtml
+++ b/templates/social/social_page.xhtml
@@ -121,13 +121,13 @@
                     Blurblogs
                 </div>
             </a>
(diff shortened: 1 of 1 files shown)


samuelclay:
Not ideal, but I'm fine with this. Better than the alternative of having a mismatched cert.

stevenatwork:
I'd like to fix this issue on another project: https://github.com/EFForg/https-everywhere/issues/1327#issuecomment-180097223

Is there a timeframe for when newsblur.com will be updated to the latest code?

PR #878, opened 2016-02-09
api: add mark_story_hashes_as_read

Zaspire:


samuelclay:
Thanks for adding this!

PR #879, opened 2016-02-12
Android: Hyperuser Fixes, Part 2

dosiecki:
Tweaking the new and improved story fetch lifecycle to have all of the behaviours we expect, like:

* only fetch a set of pages of stories for a given feed/folder once per high-level feed sync
* have reasonable timeouts
* story lists shouldn't be jumpy
* dont' show debug status lines when debug isn't enabled
* don't crash

I also upgraded our HTTP libs, since not only have there been some nice speed increases in them, there are also a few security vulns to patch (none directly affect us, but GOOG might start flagging APKs that use out-of-date ones).  Updating to the most modern version also opens the door for us to start doing interruptable IO in a slightly less messy way, should we deem it necessary.


samuelclay:
Excellent, this ready for beta?

samuelclay:
Well, it'll have to be on Tuesday as I'm heading to Reno/Tahoe for my birthday this weekend. 

Apart from adding the two new libs (okhttp-3.1.2 and okio-1.6.0), should I be changing anything else?

samuelclay:
Anyway, it's compiling fine and looking great to me. Feels faster than the first PR as well. I'm :+1:.

dosiecki:
Any time we touch libraries, I'd love it if we could all Alpha this ourselves for a few days before going public Beta.  I'll be intensely using it all weekend, for sure.

Shouldn't need to change anything other than the usual version strings, unless your development environment got confused by the library update.  I always rename files, just to make it as unambiguous as possible, but some still stumble!

manderson23:
All good for me over the weekend.

PR #883, opened 2016-02-29
Android: Beta Bugfixes

dosiecki:
Fixes for a few issues discovered in the current Beta.


samuelclay:
I'm going to be sending this out as 4.8.0b2, but I can't beta test as I'm in NYC all week and don't have my android devices on me.

PR #890, opened 2016-02-29
Android: Fix newsletter text not wrapping

manderson23:
As discussed via e-mail this CSS change overrides min-width so newsletter text can wrap properly.

Tested across various other feeds with no obvious regressions.

### DIFF
--- a/clients/android/NewsBlur/assets/reading.css
+++ b/clients/android/NewsBlur/assets/reading.css
@@ -21,13 +21,14 @@ p, div, table {
 	line-height: 1.5em;
 }
 
-p, a, table, video, embed, object, iframe, div, figure, dl, dt {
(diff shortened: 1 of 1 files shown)


samuelclay:
Woah, this works? I'll give it a try tomorrow. Perhaps we want to only set this on feed.is_newsletter? Well, min-width doesn't seem to have any effect on other stories, so maybe it's fine for all feeds. Nice thinking!

dosiecki:
All my newsletters appear to be plaintext-ish, so I haven't seen the problem, but I spot-checked some of the various other elements touched by the change, and so far everything looks good!

manderson23:
There was one particular newsletter where a center element had a min-width set to larger that the width of the screen in portrait.

All my newsletters seem to be HTML heavy particularly with their use of tables for layout. This means they don't look great when using the dark theme so I'm going to look into that next.

PR #892, opened 2016-03-05
Android: Dark theme improvements for newsletters.

manderson23:
Improves readability of HTML heavy newsletters in dark theme.

### DIFF
--- a/clients/android/NewsBlur/assets/dark_reading.css
+++ b/clients/android/NewsBlur/assets/dark_reading.css
@@ -1,22 +1,24 @@
-body, span {
+body, span, table, td, p, div {
 	background-color: #1A1A1A !important;
 	color: #FFF !important;
 }
 
 a, a span {
 	color: #319DC5 !important;
+	background-color: #1A1A1A !important;
 }
 
 a:visited, a:visited span {
 	color: #319DC5 !important;
+	background-color: #1A1A1A !important;
 }
 
-code {
+pre, blockquote, code {
     background-color: #4C4C4C;
 }
 
-pre, blockquote {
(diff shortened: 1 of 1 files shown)


clients/android/NewsBlur/assets/dark_reading.css
```diff
@@ -1,22 +1,24 @@
...
 }
 
-code {
+pre, blockquote, code {
     background-color: #4C4C4C;
 }
 
-pre, blockquote {
-    background-color: #4C4C4C;
+blockquote > p, pre > span, code span {
```
samuelclay: What about just `pre`?

manderson23: pre is now handled in the previous selector.

samuelclay: Doh, I should have noticed that. Just got back form NYC and I'm at half-functioning.

samuelclay: Can you give me examples of the newsletters fixed by these selectors? I need to also implement this on iOS and possibly web.

manderson23: http://www.nytimes.com/newsletters/what-were-reading?nlid=72249677 - still
needs a few tweaks. Title text is still black for example.

http://softwareleadweekly.com/

Pocket hits - not sure there is a link without being signed in

http://www.mountaingoatsoftware.com/ - weekly e-mail tips

http://www.longreads.com/ - weekly e-mail. This is the one the had inline
!important do doesn't look good at all for now.

PR #894, opened 2016-03-15
Android: More dark theme improvements.

manderson23:
With !important stripped these changes make the HTML based newsletters look reasonable.

### DIFF
--- a/clients/android/NewsBlur/assets/dark_reading.css
+++ b/clients/android/NewsBlur/assets/dark_reading.css
@@ -1,4 +1,4 @@
-body, span, table, td, p, div {
+body, span, table, td, p, div, li {
 	background-color: #1A1A1A !important;
 	color: #FFF !important;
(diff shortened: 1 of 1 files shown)


samuelclay:
Are we ready for wide release?

samuelclay:
By the way, do you see any width issues? Many newsletters are 600px.

dosiecki:
> Are we ready for wide release?

I would suggest we hold off. I just last night found a fairly frustrating behaviour while reading on the subway. With very unlucky timing, an unintended blocking of the UI loop can occur if a network call goes crazy latent.  Stories in the UI will freeze and the effect persists until timeout. 

I know the exact reason, just need to get in front of a real keyboard to fix when I arrive home tonight.

manderson23:
After adding the min-width override I've not seen any issues with
newsletters I read.

PR #895, opened 2016-03-16
Android: Fix loading of stories in bad network conditions

dosiecki:
Fixing a bug where stories (even prefetched/offline) can fail to load or may load into the wrong views when the device's network state is blackholed (not offline, but experiencing extreme timeouts).


samuelclay:
Beta release? Or time for prod?

dosiecki:
Beta for sure, Prod as soon as we're all satisfied with our usual tests on different accounts, themes, etc.

PR #896, opened 2016-03-18
Android: Fix on-upgrade crash.

dosiecki:
Very quick patch for an on-upgrade crash that is blocking release.  Will address a proper fix for this during the next cycle.


dosiecki:
Save for the crash fixed by this change, I think 4.8.0 is looking stable for release. (we didn't push this fix to Beta, correct?)

samuelclay:
Yeah, I didn't bother putting out a beta for it. Ok, I'll release soon. I think I'm going to time it for next week on Monday. That's the day after Easter and people typically read more (and not work) immediately after holidays.

Issue #898, opened 2016-03-20
Android: Add Saved Stories segmented control to All/Unread/Focus


samuelclay:
Saved stories by feed, showing the saved count per feed in blue (same as the unread counts for saved stories). This exists on the web.

Adding `read_filter=starred` will get you the stories, per-feed. 

The counts are in `starred_counts` from /reader/feeds?flat=true. It is two parts merged:

```
...
{
    "count": 1,
    "feed_address": null,
    "feed_id": 5077457,
    "tag": null
},
{
    "count": 1,
    "feed_address": null,
    "feed_id": 97100,
    "tag": null
},
{
    "count": 5,
    "feed_address": "http://www.newsblur.com/reader/starred_rss/13/db0d7a788724/art",
    "feed_id": null,
    "tag": "Art"
},
{
    "count": 3,
    "feed_address": "http://www.newsblur.com/reader/starred_rss/13/db0d7a788724/",
    "feed_id": null,
    "tag": "being human"
},
...
```

The top two are the feed counts, bottom two are the tag counts. Why I merged the two I don't know but it's easy to separate them.

dosiecki:
Hmm, this way of looking at saved stories by feed is a bit different than how I had the design in my head and how I was already coding things.  I had thus far been treating saved counts for a feed as a sort of specialised tag and showing both tags and feed groupings in the expandable tree under the all-saved-stories list item.

This made sense to me, in that it puts all of the saved stories info in one spot in the UI, lets you see saved counts by feed, and avoids the myriad evils of a modal filter UI.

Would you be willing to entertain a demo version of the UI that works in this way? I am liking the design for my own use, and it is virtually finished, once I merge this code that I started a month ago with all of the hyperuser rewrites that pushed it back.

samuelclay:
Sure I'll take a look, but why not merge the two and let feeds show their own saved story count?

dosiecki:
*(Bias Disclaimer: I find modal interfaces deeply confusing for users and Android has a really bad history with them; users often end up saying "my stories/emails/photos are missing!" when in reality they were just in the wrong mode.)*

[...]

PR #902, opened 2016-03-31
Android: Saved Stories by Feed

dosiecki:
Implementing reading of saved stories as split by feed with the UI arrangement discussed in #898.  Note that this is only the functional bits and the general UI structure, not the new icons or segment control formatting.  Those will come in the next PR that will have lots of UI cleanup and restyling.


yardenac:
Thanks so much for doing this! I'd been waiting for years, can't wait to use it.

PR #905, opened 2016-04-07
Android: Housekeeping

dosiecki:
Attempting to fix up our Android build environment to be more standard, flexible, and to satisfy automated F-Droid builds that recently broke.  Also updated to newer versions of some of our dependencies.

*Strongly recommend doing a `clean` or `nuke` in your build environment before attemping to build after these changes.*


samuelclay:
Error:Gradle version 2.2 is required. Current version is 1.12. If using the gradle wrapper, try editing the distributionUrl in /Users/sclay/projects/newsblur/clients/android/NewsBlur/gradle/wrapper/gradle-wrapper.properties to gradle-2.2-all.zip.

Please fix the project's Gradle settings.
<a href="openGradleSettings">Gradle settings</a>

dosiecki:
I would suggest getting the latest gradle.  1.12 is quite out of date.  I have tested with 2.6 and above with no problems.

samuelclay:
Now getting a Proguard error. I swear I'm googling these before I post it here.

Error:(15, 0) Build script error, unsupported Gradle DSL method found: 'runProguard()'!

Possible causes could be:  
  - you are using Gradle version where the method is absent (<a href="open_gradle_settings">Fix Gradle settings</a>)
  - you didn't apply Gradle plugin which provides the method (<a href="apply_gradle_plugin">Apply Gradle plugin</a>)
  - or there is a mistake in a build script (<a href="goto_source">Goto source</a>)

dosiecki:
What build environment are you using?

samuelclay:
It really doesn't want to work now. I'm following the building directions with Android Studio and no modules are being imported when I start from a clean repo and import clients/android/NewsBlur.

dosiecki:
Hmm, I know AS uses gradle internally, so it is possible that our
standalone configuration is confusing it. I will have to install AS and
play around to see how to smooth out compatibility. In the meantime, you
should be able to build with gradle from a shell or just
temporarily/locally rm the `build.gradle` config file and AS will function
as before.

dosiecki:
I am a bit stumped.  I created a new AS project (in AS v1.5.1) pointed at the Android client root directory for import, and it built perfectly with even fewer steps than in the past.  If anything, the new gradle config file made the import go much more smoothly (there was one question about letting AS manage the gradle environment version, but that was it).

(This was on GNU/Linux. I also borrowed a Windows PC and tried the project with the newest AS v2.0, and though the install process was so awful that it broke gradle, it fell back to the legacy build method just fine and built the APK.)

samuelclay:
Oh dear lord I figured it out. I have no idea how this happened, but I was running a 0.8.4 BETA. Forget the fact that as of yesterday AS was version 1.6, but I was running a *beta* of some old version. I'm just going to download AS 2.0 and hope for the best.

samuelclay:
I wonder if I would have figured that out if AS 2.0 wasn't released today.

samuelclay:
Excellent, all's well again.

Could you make the Saved icon in the feed list's all/unread/focus/saved control a bit larger? We just implemented this on iOS and had the same issue.

samuelclay:
Also, I'm getting 0's for unread counts.

samuelclay:
By the way, that footer's looking nice! Can we turn those font sizes down just a touch? Also, lets use the iOS icons. I don't think color belongs in that toolbar.

dosiecki:
> By the way, that footer's looking nice! Can we turn those font sizes down just a touch? Also, lets use the iOS icons. I don't think color belongs in that toolbar.

Created #908 to track (I swear I had an issue for that, but it must have been closed)

Issue #908, opened 2016-04-14
Android: Visual Update for State Toggle


dosiecki:
Update the all/unread/focus/saved toggle that appears under the feed list.  Includes updating to new icons, better formatting, and tweaks to fit on all screen sizes.

dosiecki:
implemented in #912

manderson23:
I've noticed an inconsistency between the colours on the circles in the navbar and the colours used in the toggle now.

The navbar has a grey circle for unread and a green circle for focus. Whereas the toggle now uses grey for focus instead of green.

Should the circles be consistent?

samuelclay:
That's OK. It comes from the fact that on iOS, toolbars can't have color. They just need to be different. Also, I think it looks better in grey than it does using the unread bullet color.

dosiecki:
Though I don't mind the look (the aesthetic of assets going monochrome when
in the toolbar is pretty consistent on many modern platforms), I have had
this same question in my head. It looks like they are similarly mismatched
on iOS from the screenshots I have seen, and Web still uses green / grey.
This would be rather simple to switch out, we just need to decide which is
desirable.

samuelclay:
The web should probably change.

PR #912, opened 2016-05-03
Android: Housekeeping and new Intel State Toggle

dosiecki:
The primary functional change here is the implementation of #908: a big visual update to the long-neglected state toggle widget.  Not only did it previously fail to match either the NewsBlur web or iOS UIs, it also failed to fit on smaller device screens since the addition of the 4th (saved) state and grossly violated pretty much every Android UX guideline.  I took a somewhat creative approach to satisfying all of these requirements at the same time.  The general look and feel matches the newest iOS style, but translated into the design language of the Android platform.  I solved the space/clipping problem by making the buttons collapse when not selected. (this is best explained by playing with the toggle)  All four buttons now nicely fit on even the smallest device we support.

The secondary set of changes here is a ton of housekeeping that I wanted to get done before any more functional changes.  This included getting rid of dozens of lint errors, compile warnings, and deprecated API touchpoints.  Also fixed some various warnings that AS and gradle liked to throw during the new and improved build processes.  Finally, removed some API usages that were problematic on the most recent Android "N" version and causing trouble there.  I think this is the first time in over two years that it is possible to do a zero-warnings release build!


samuelclay:
Looks great! I'd be happy to ship this out as a release. Beta first? I don't see a need, unless you do.

dosiecki:
I am still running through my test gamut (especially since this contains fixes specific to particular platform versions), but will ping with an OK-GO when done.

PR #914, opened 2016-05-11
Android: Fix on-upgrade crashes

dosiecki:
Fixing some one-time on-upgrade crashes that we were seeing with the last few Betas.


dosiecki:
I have finished up my testing for this release. We have one rather obvious
visual glitch on devices running Android 4.0.X, but << 5% of our active
users are on devices that old, and everything is still functional.

If y'all are happy with this latest build, we could launch now, or if we
want to fix the 4.0.X glitch, I should be able to proc a fix by Monday.

samuelclay:
Ok, happy to wait.

PR #917, opened 2016-05-16
Android: Fix UI glitches on API 14 and 15 devices

dosiecki:
working around some ugly things resulting from buggy early implementations of the UI framework.

### DIFF
--- a/clients/android/NewsBlur/res/drawable-nodpi/toggle_set_border.xml
+++ b/clients/android/NewsBlur/res/drawable-nodpi/toggle_set_border.xml
@@ -3,6 +3,8 @@
 
     <item>
         <shape>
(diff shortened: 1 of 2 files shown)


samuelclay:
Sounds like a fine compromise.

dosiecki:
From here on out I'm not going to over-spend resources on API 14/15 issues unless they impact stability.  I fully expect we have <5% of active users on those versions and we will consider dropping support for 14/15 soon after we add support for API 24. (older versions will keep working, but updates will focus on devices made in the last 5 years)

I am more than ready to do a Beta from where we are (just 2 days to check that we got rid of on-upgrade crashes), unless y'all have found any issues.

samuelclay:
Feel free to drop support earlier. Right now we only support iOS 8 and iOS 9.

Pushed out as v4.9.0b1.

PR #918, opened 2016-05-19
Android: Fix On-Upgrade Crashes

dosiecki:
Finally gave in and implemented some more strict on-upgrade checking.  Newer devices were launching the UI so quickly, sometimes the app would crash because the UI was trying to access the DB before it was upgraded.

There is now a splash screen when the app is opened the first time after install or upgrade while the check is occuring.  Total time from app start to the UI being displayed in full has actually stayed the same or gone down (time in the spash is == the time the UI used to pause with the "Tidying up" message), though being forced to use a spash might make it appear a bit slower due to some unfortunate cognitive biases.  Can't think of any more reliable way to finally squash the on-upgrade crashes, though.

Also disabled debug logging, since I expect this will be our final release candidate if all goes smoothly.

### DIFF
--- a/clients/android/NewsBlur/AndroidManifest.xml
+++ b/clients/android/NewsBlur/AndroidManifest.xml
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.newsblur"
     android:versionCode="125"
-    android:versionName="4.8.1b2" >
+    android:versionName="4.9.0b2" >
 
     <uses-sdk
         android:minSdkVersion="14"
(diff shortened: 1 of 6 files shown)


samuelclay:
Pushed v4.9.0b2 to play store.

dosiecki:
I've tested the upgrade path of the last 3 major versions (and half a dozen minor versions and betas) to this build, and all were clean!

How is this build looking for y'all?  Zero crashes on the market so far, so unless anyone spots a regression, I am green for prod.

samuelclay:
Sorry to keep you waiting, I'm holding off on this release until next week, as I'm going camping for memorial day weekend and will be offline until Tuesday.

PR #923, opened 2016-06-04
iOS: Clean up activity handling

nriley:
I couldn't figure out why a UIActivityItemProvider subclass also implements UIActivityItemSource (working around an old bug?) so I picked one.  I re-added the URL as a separate activity item; it works well in everything I've tested. Particularly when sharing to Twitter, Facebook, Reminders and Notes it looks better include the URL separately.  It also eliminates an extra newline that was showing up when sharing to OmniFocus.

Another motivation was fixing sharing to Reading List (see https://getsatisfaction.com/newsblur/topics/send_to_reading_list_not_working_in_ios_apps).

Note: tested on iOS 9 only (don't have anything earlier installed anywhere).


samuelclay:
Oh shoot, sorry I missed this. I should've merged this in pronto. Just saw you discussing this in the support forum. 

I'll test for a couple days and then submit as an update.

PR #928, opened 2016-06-24
Switch to filter instead of opacity

jryans:
In Firefox, opacity causes the background color underneath the container to pop through. By using filter instead, only the elements on top of background are affected.

In Chrome, they both have the same effect.


samuelclay:
Sounds good to me. Thanks for taking the time to fix this!

Issue #903, opened 2016-04-05
Android: Missing fleuron at bottom of story titles list


samuelclay:
What it looks like now: 
![2016-04-05 03 47 40](https://cloud.githubusercontent.com/assets/44229/14270799/96b97b10-faa6-11e5-86bd-736a41119104.png)

Here's the fleuron:
![2016-04-04 20 49 13](https://cloud.githubusercontent.com/assets/44229/14270813/ba2df24c-faa6-11e5-8c78-beebbf56b63c.png)

samuelclay:
While here, let's change the background color to match iOS. The white is too bright.

Also, do we have a ticket for the story image previews?

dosiecki:
> Also, do we have a ticket for the story image previews? 

Created #904

dosiecki:
Just to be clear, by 'fleuron' do you mean the little monochrome NewsBlur logo that demarks the end of a story list?

samuelclay:
Correct. It's labeledl fleuron.png in the repo.

samuelclay:
By the way, I'd love to see story thumbnails next. I can see them in this screenshot and I'd think they'd look great on Android as well. Obviously needs a preference for those who want the extra horizontal space for text.

dosiecki:
>I'm not seeing these at the bottom of story title lists. I'm reading by folder and the last story doesn't budge from the bottom of the screen. 

Very odd.  Both the resource and the code to use it are in HEAD:

https://github.com/samuelclay/NewsBlur/blob/master/clients/android/NewsBlur/res/drawable/fleuron.png
https://github.com/samuelclay/NewsBlur/blob/master/clients/android/NewsBlur/src/com/newsblur/fragment/ItemListFragment.java#L173

Does your build process need an additional nudge to pick up new resource artifacts?  Maybe try a full nuke and rebuild?

manderson23:
I've just pulled these changes and noticed that the bottom of the fleuron circle seems to be clipped?

dosiecki:
> I've just pulled these changes and noticed that the bottom of the fleuron circle seems to be clipped? 

Yeah, I meant to ask about that.  The source PNG from the iOS tree was like that.

samuelclay:
Looks fine to me on my S5. Releasing as 4.10.0b1.

manderson23:
On Tue, 5 Jul 2016 21:48 Samuel Clay, <notifications@github.com> wrote:

[...]

Issue #900, opened 2016-03-21
Android: Forgot Password flow


samuelclay:
Just open a web view with: http://www.newsblur.com/folder_rss/forgot_password

dosiecki:
done in #930

Issue #861, opened 2016-01-18
Android: Adjustable Text Size for Lists


dosiecki:
The range of DPI values we're seeing in the wild across devices is getting wider and wider, making it hard to pick a correct font size for feed lists and story lists and story metadata.  We've now got compelling examples where our chosen size is both too small to read and too large to fit much on screen.  We should consider adding a text size chooser for the rest of the app, not just story reading.

samuelclay:
Yes I agree. I have two sizers for web and iOS: feed and story list, and story detail.

dosiecki:
done in #930

Issue #837, opened 2015-11-30
Android: Keypad UX


dosiecki:
Was pointed out to me that some of the Android UI can't be completely navigated and used via keyboard/keypad due to focus and selection display issues.  This is not only a nice-to-have feature, this is the interface that is used by accessibility apps such as eye trackers and integrated joysticks for disabled users.

A surprising amount of the UI does already work if you focus via tap or mouse, we just need to iron out how to move from pane to pane in a sensible way.

dosiecki:
Also discovered today this this is a blocker for supporting devices with hardware keys to scroll, since using the hardware key unavoidably pops us into key UX mode, which causes some ugly movement issues.

carbin:
The BlackBerry Priv has a physical keyboard with capacitive keys so you can swipe up and down on the keyboard to scroll. In the NewsBlur app this works for the feed list but not for the story view. I think it's related to this issue.

dosiecki:
Yeah, I should be clear - this is known to be broken in story view on all devices, and only partially working in the story list and feed list.  Just to set some expectations, this is going to be more of a long-term goal than a quick fix.  The way Android handles keyboard or keypad inputs is changing rapidly with newer versions of the OS and the story view in particular is somewhat out of our control, as we let the OS render story content and it steals keyboard control from us.  I have some ideas around how to make this work a bit better when we refresh the button UI on the story view.  I will ping this issue/task whenever there are changes that might merit some new testing and evaluation.

PR #930, opened 2016-06-30
Android: UI Improvements

dosiecki:
This change goes after a good number of small UI tweaks for which we've had open issues.  In particular:

* added fleurons to completed story lists (#903)
* fixed list background colours (per comments in #903)
* adjustable font sizes for story lists (#861)
* forgot-password workflow (#900)
* minor fixes to focus pathing (#837)

I also did a pass over some mismatched or slightly-off theme colours and made them more closely match the iOS UI or the web UI or both (where possible), vaguely guided by the Android UX guidelines (again, where possible).  I would say that about 75% of the app now has a nice, clean, consistent palate; I'll be finishing off the remaining artifacts in the next pull.

Finally, a handfull of bux fixes.  Also bit the bullet and cut off update support for the oldest ~4% of Android devices out there that were costing us a ton of time in testing and support.


samuelclay:
Gonna push this out as a beta.

samuelclay:
There is something bizarre happening, where when I go into a feed, stories from other feeds show up and don't go away.

dosiecki:
Is your device under duress of any sort or does the problem go away after letting it sit for a moment? Try swiping the app closed and seeing if that changes things. Also, device / platform version info?

dosiecki:
Went back and confirmed that nothing about sync has changed since the last release, so if there are any issues, they aren't regressions.  One other diagnostic in addition to the usual feedbacklink data - was your Android device turned on for the first time in a while when the test was performed?

samuelclay:
Yes it was. I unplugged it a couple weeks ago and forgot to charge it. So it was just turning back on, although I let it charge for an hour.

samuelclay:
Samsung Galaxy S5

Issue #936, opened 2016-08-03
Android: Rename "Send Full Story" or use Text view


samuelclay:
See https://getsatisfaction.com/newsblur/topics/android-send-full-story-to-doesnt-send-full-story. I'd be happy with either, but I agree that it shouldn't say Full Story unless it's the Text view. Perhaps make this conditional based on whether or not the Text view has been loaded for the story. Full Story when yes, "Story" when just the regular feed content.

dosiecki:
fixed in #938

Issue #931, opened 2016-07-05
Android: Don't show "Saved Stories 0"


samuelclay:
While the feed list is loading, there's a blue 0 unread count on Saved Stories. If the number is 0 (or below), just hide the counter.

dosiecki:
fixed in #938

Issue #922, opened 2016-06-01
Android: Improve Icon Cache Cleanup


dosiecki:
The newest Android / Google Play Store framework has a feature that suggests to users that apps should be uninstalled if their use of device storage is too high.  This is probably a good time to get our ducks in a row w.r.t. how we cache icons.

Though they are very small and the dataset changes very slowly, we essentially never clean up the favicon/profilepic caches.  We should refit those caches with the nice, new cleanup framework that we implemented for the offline image cache last year.

dosiecki:
fixed in #938

Issue #921, opened 2016-05-29
Android: Custom server preference


Nekmo:
I just got NewsBlur install on my server but I can not see how to define the url of my server in Android application.

dosiecki:
The Android app can be found in this repo, here:  https://github.com/samuelclay/NewsBlur/tree/master/clients/android/NewsBlur

The file where the API endpoint is defined can be found here:  https://github.com/samuelclay/NewsBlur/blob/master/clients/android/NewsBlur/src/com/newsblur/network/APIConstants.java

Nekmo:
Do I really need to modify the application for this? Is this a bad joke?

samuelclay:
We should add a preference for defining a Custom Server. We have this on iOS.

cedricpim:
Hi @samuelclay, any ideas on when this might be implemented? Or is there any android apps that would support NewsBlur with a custom server?

samuelclay:
Hey @dosiecki, would this be easy to add?

samuelclay:
I see referenced on #938 that this will have a login screen element. Don't bother adding a custom server to the login screen. Just add a preference. The user can login to NewsBlur first to make the domain switch.

dosiecki:
On 29 December 2016 15:49:17 GMT-08:00, Samuel Clay <notifications@github.com> wrote:
>I see referenced on #938 that this will have a login screen element.
>Don't bother adding a custom server to the login screen. Just add a
>preference. The user can login to NewsBlur first to make the domain
>switch.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/samuelclay/NewsBlur/issues/921#issuecomment-269711629

Will the cookie received from the official server work with others? I figured that the login call that retrieves it needed to be to the same endpoint that would be hit by calls after changing it.

samuelclay:
It's a local preference, so it would discard the cookie when connecting to a new server. I see your point in that if you can't login on your server then you're SOL because you can no longer access preferences. But they can reinstall the app. Not the best fix, but for somebody who is running their own server weren't not going to be doing custom UI work to support that beyond exposing the preference. Not yet anyhow.

dosiecki:
[...]

PR #938, opened 2016-08-04
Android: Fixes and UI Tweaks

dosiecki:
Several bugfixes for new and outstanding issues, including #931, #922, #936, and the stale stories bug.

Also includes the second half of the UI tweaks that were started in #930.  Nearly all widgets have been updated to be agreeable to the platform design spec while still using the house (web/iOS) style if at all possible.  Many icons redone from vector/PSD in significantly higher resolution.  Some old raster elements converted to vector.  UI cleanup is now complete, except for the login screen, which will best be addressed when we do #921.


samuelclay:
"All Stories" title in story list but "ALL STORIES" title in story detail. Can we keep standard case for title in story detail?

Otherwise, looks good! I'm going to put this out as 1.11 beta 1.

dosiecki:
> "All Stories" title in story list but "ALL STORIES" title in story detail. Can we keep standard case for title in story detail?

Hah, good catch.  I have somehow stared at those two strings every day for over a year and never noticed they were off.  Pull incoming.

Issue #904, opened 2016-04-05
Android: Content Image Previews in Story List


dosiecki:
The Android app should support the optional story image thumbnail when story content previews are enabled.

dosiecki:
Update: the work done in 13d3fd5 directly supports finally going after this.

dosiecki:
implemented in #942

PR #942, opened 2016-08-19
Android: Story Thumbnails

dosiecki:
Implemented #904 - thumbnails in story lists.  This took a bit of refactoring, since loading up that many images of unknown sizes and network origin very quickly is enough to bring many Android devices to their limits of memory, CPU, and IO.  Finally came up with a solution that works without interaction lag on even my 4-year-old phone, though.  Includes support for offline reading.  Option is disabled by default, since using it could lead to some very high and unexpected data usage for mobile users.

Also took the time to really clean up our image caching framework (offline reading, thumbnails, and icons) now that we are doing so much with it.  Gone should be the days of leaked storage and wild memory use numbers.

Note that this includes a DB schema bump and also some dependency bumps.  A full clean and build may be required.


samuelclay:
Looking good! Few changes I'd like to see:

- Images should go top to bottom edges. See screenshot: 
- Stories without preview images should extend all the way to the right. I think they check for the `image_url` key.
- Let's default to it on. Users can turn it off.

samuelclay:
While we're here, add a single pixel white line at the top of each story. Notice you can see it in the iOS screenshot.

samuelclay:
Lastly, change the toolbar (top and bottom) color to #E4E7E1 to match iOS and the web.

dosiecki:
> Images should go top to bottom edges. See screenshot. 

On my device, images are clipped to square so that they extend fully into the thumbnail viewport and that port goes all the way to the dividing line between stories, save 1 dip to account for the fact that some platforms get view sizes off by 1 and overlap would look really grungy.

> Stories without preview images should extend all the way to the right.

I tried this and it looked pretty bad for feeds with a mix of having and not having thumbnails, since the right edge of the story titles becomes all mis-aligned (as seen in your screenshot).  Can re-enable this if you like.

> Let's default to it on. Users can turn it off.

I ran a quick test where I did about half an hour of my usual reading habits and tracked how much extra data the thumbnails used.  It came out to about 50MB extra data just in that short time, even *with* prefetch turned on (it likely would have been 5-10x higher without prefetch).  This only cost me $1.50 on my data plan, but had I been roaming or overseas, it could have been a nasty $20 surprise.  I would hesitate to default this to on unless we pop up a warning of some sort.  I think the core issue here is that those tiny thumbnails are clipped down on the client side, often pulling huge full-screen sized sources down over the network, hence the huge cost.  Is there any chance we could ever switch to processing the thumbnails on the server side?

> While we're here, add a single pixel white line at the top of each story. Notice you can see it in the iOS screenshot.

We request a 1dip divider between stories.  It shows up for me on 6 out of 7 of my test devices and the one where it doesn't is running an intentionally very old framework for bug chasing.  If you're missing the divider on any current device, send me your debug info and I will take a look to see if we can detect it and work around.

> Lastly, change the toolbar (top and bottom) color to #E4E7E1 to match iOS and the web.

I actually spent several hours this summer trying to purge that colour from the app.  I didn't realize it was fully intentional, but on most oLED screens, it renders as pea-green.  The current colour is just desaturated so it looks gray like I see on the web on my monitor (though I will admit that now that it has been pointed out, I do notice a sort of yellow-ish tinge on the web).  Is there some compelling reason not to keep these monochrome like most system controls?

samuelclay:
Yes please re-enable ragged right. I know it's mis-aligned, but it shows more information at a reasonable cost.

Images on the server-side would handle another big issue: http-over-https. I would love to serve images, though it's hard to justify the added cost. I don't see this happening anytime soon. I know there are cloud services that can effectively do this for you, but they still cost a pretty penny. Ben from Feedbin told me that it doesn't cost him much, but I'm still wary of going through the trouble.

I'm fine with 50MB over a half hour. That's quite a long time, and I'm not aware of data plans in the U.S., known for having some of the world's most expensive mobile plans, that would cost $1.50 for 50MB unless they were roaming. We're optimizing for people who read news on their phone. They know it costs data, that's what they're going for. Let's turn it on and simply let people turn it off. 

On iOS we have two 1px dividers, white on top and grey on the bottom. It creates a consistent shadow effect.

Heh, "pea green" is NewsBlur's color! It's intended to look faintly green. It's to balance the yellow and orange of the logo. Plus I can recognize NewsBlur versus every other app just by glancing at the toolbar color, web and mobile.

dosiecki:
> Yes please re-enable ragged right.

Will do.

> I'm fine with 50MB over a half hour

Will change the default to enabled.

> On iOS we have two 1px dividers, white on top and grey on the bottom. It creates a consistent shadow effect

Ahh, gotcha.  Will try and tweak the containers to provide something similar.

> Heh, "pea green" is NewsBlur's color! It's intended to look faintly green. It's to balance the yellow and orange of the logo. Plus I can recognize NewsBlur versus every other app just by glancing at the toolbar color, web and mobile.

Interesting!  I somehow always saw it as grey up until testing the app on a device with an awful pentile screen and it was so green I thought I was going crazy.  Will do a run back through the default/light theme to ensure it sticks.

PR #943, opened 2016-08-24
Fix stories intermittently not displaying; clean up story loading

nriley:
This should fix the blank-story issue — that's been bugging me for weeks, but seeing complaints about it on Get Satisfaction tonight got me sufficiently motivated to fix it.

There's a lot of stuff in here I cleaned up that I don't completely understand (and its commit history is not terribly helpful). Maybe some of it is for old iOS versions? Please review and/or test!

### DIFF
--- a/clients/ios/Classes/StoryDetailViewController.m
+++ b/clients/ios/Classes/StoryDetailViewController.m
@@ -35,8 +35,7 @@
 
 @interface StoryDetailViewController ()
 
-@property (nonatomic, strong) NSString *loadingHTML;
-@property (nonatomic, strong) NSURL *loadingURL;
+@property (nonatomic, strong) NSString *fullStoryHTML;
 
 @end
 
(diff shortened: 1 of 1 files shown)


samuelclay:
Nice work! I'm going to merge this and we'll see if it causes issues. I'm heading to Burning Man tomorrow so I won't be able to confirm this, but I'm loading this version up on my phone so I'll have it ready. Of course then my phone will be turned off for a week and a half.

nriley:
I have discovered at least one regression, which is that the hover bubble suppression is not working properly (some injected JavaScript). I should be able to work on this in the next day or so.

nriley:
OK, now I can't reproduce this.  What was going on was that when tapping and holding on an image, the magnification bubble (callout) was appearing briefly before the expected popup.  I did change where and how often this JavaScript gets injected so there could be a race here, so let me know if you see it and I'll try harder to reproduce it!

samuelclay:
Ok, I'm including this commit as part of 6.1.0, which I'll submit tomorrow.

Issue #946, opened 2016-08-26
self-hosted documentation


creactiviti:
I'm racking my brains for a couple of days trying to get NewsBlur installed on a local VM and had failed miserably so far. 

Any way we can get a step-by-step installation guide?

jrottenberg:
You should use docker installation, it's pretty straightforward with docker-compose.

creactiviti:
thank you @jrottenberg. I'm running into issues with that as well though. I think that would work great for the application itself but not for the persistent pieces (i.e. Postgres/MySQL, Redis etc.)

samuelclay:
Just waiting on #947 to fix some formatting and we'll have that merged in.

remram44:
I just started NewsBlur via docker-compose, however I'm noticing that feeds update really slowly (one feed every couple minutes... so 3 hours to fetch 100 feeds). Moreover clicking on "insta-fetch" shows a rate-limited message.

Is there a way to tweak the refresh delay/rate limit somewhere? Am I hitting another bug?

samuelclay:
So the fetching that's happening there is when the client sees that feeds aren't being updated and updates them itself (on the app servers). It essential,lly Insta-fetches the feeds since they aren't being updated in the background.

You need to run the celery workers. I would just have them run every single queue and not split them up. If you omit the queue name, celery will just handle them all.

You also need to setup celery's celery beat, which performs scheduled tasks. Those tasks are what fill the feed queue and then send the stale feeds out to the task queues.

Lich4r:
Hi,

How did you manage to install newsblur with docker ?

I've read the docker installation but I guess I'm missing something.

PR #947, opened 2016-08-29
Docker fix

jrottenberg:
Maybe some hope for #946

It's better than the first version, fresh docker-compose up should give you a working newsblur environment.


samuelclay:
I would love to merge this but can you undo all the spacing changes? I just want the real diff. It looks like your text editor cleaned everything up. That would be fine, but I would want that as a separate PR for sure.

jrottenberg:
np, I'll strip the formatting.

jrottenberg:
ok rebased master and cleaned up the commits, the intent is more clear, let me know if anything is not explicit.

utils/feed_fetcher.py
```diff
@@ -34,7 +34,7 @@
 from django.utils.encoding import smart_unicode
 from utils import json_functions as json
 from celery.exceptions import SoftTimeLimitExceeded
-from vendor import tweepy
```
samuelclay: Can you undo these two? Tweepy is explicitly vendorized because of some changes that I needed to make.

samuelclay: Aww geez, you're absolutely right. Thanks for catching this!
screen shot 2016-09-16 at sep 16 2 39 11 pm

remram44:
I'm confused -- is the task server still missing from docker-compose?

jrottenberg:
Not sure to follow you, the main service is labelled **newsblur** listening on port 8000. For the newsblur service it just starts the UI (gunicorn), I don't start any worker, it's doable but at the moment it's not clear what workers (see config/supervisor_*) are mandatory to be able to have a service running. I've defaulted to the simple webservice.

remram44:
This is what I'm getting:
screenshot

I've added some celery workers but somehow they don't update the feeds? Worker log is full of:
```
celery_1         | [2016-10-01 02:07:18,138: INFO/MainProcess] Received task: task-feeds[af60f3d2-e960-46ba-93d2-e399bf351c48]
celery_1         | [Oct 01 02:07:18]  ---> No feeds to queue! Exiting...
```

I'm not sure how all the queues & tasks work, and the difference between feeds that are "tasked"/"queued"/"scheduled". My changes are here: https://github.com/remram44/NewsBlur/compare/master...docker-worker

remram44:
Actually it seems like it's updating now... It looks like some task needs to go over the feeds and schedule them for updating? not sure what's this about.

remram44:
Continuing my investigation, the feeds I upload as OPML don't seem to update until the `task-broken-feeds` happens and schedules them. This looks like a bug in either the timing of the task (only happens once every 6 hours) or the OPML importer (why are they added in a state that won't get them refreshed until the task fixes it?)

samuelclay:
I answered this a bit in #946:

> So the fetching that's happening there is when the client sees that feeds aren't being updated and updates them itself (on the app servers). It essential,lly Insta-fetches the feeds since they aren't being updated in the background.

> You need to run the celery workers. I would just have them run every single queue and not split them up. If you omit the queue name, celery will just handle them all.

> You also need to setup celery's celery beat, which performs scheduled tasks. Those tasks are what fill the feed queue and then send the stale feeds out to the task queues.

remram44:
I eventually figured that out, I'm just surprised that `broken-feeds` needs to go over them for them to start updating after an OPML import...?

I might open a pull request at some point, though realistically paying for your hosted service might be a better option, since this is too heavy for self-hosting on my home server 😅

samuelclay:
Yeah, it is a lot of work to run. It's doing a ton of things, and that's assuming you've got search, socket.io/node, and all the queues set up. I'm not sure why the feeds didn't have next_scheduled_update times filled out correctly (that's what broken-feeds checks for), but that would explain why there's always a handful of feeds that fall through.

PR #958, opened 2016-10-24
HTTP feed fetcher improvements

da2x:
- Send an Accept header asking for feed formats (instead of the current `*/*`)
- Don’t send A-IM without an If-None-Match or If-Modified-Since.
- Fixed Train-Case in “Accept-Encoding” header.

### DIFF
--- a/utils/feed_fetcher.py
+++ b/utils/feed_fetcher.py
@@ -123,8 +123,8 @@ class FetchFeed:
             try:
                 headers = {
                     'User-Agent': USER_AGENT,
-                    'Accept-encoding': 'gzip, deflate',
-                    'A-IM': 'feed',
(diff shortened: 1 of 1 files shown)


samuelclay:
LGTM. How did you figure out that these changes need to be made? Specific feeds?

da2x:
> How did you figure out that these changes need to be made? Specific feeds?

I’ve been going through caching and delta feed logic for a bunch of open source news readers. Feed subscriptions consume the most amount of bandwidth from my own website. Broken cache implementations seemed to be the root cause; and I went ahead and added delta feed support to a whole bunch of new feed readers. I’m not sure how much it will help, but it was worth a shot.

samuelclay:
Thanks, I appreciate it. I've already deployed the changes, have you noticed an improvement?

Issue #496, opened 2014-03-11
Android: Four new gestures


samuelclay:
New gestures:
- Double tap story with one finger
- Two finger double tap story
- Long press story title
- Swipe story title left
- Swipe story title right
- Long press feed title
- Swipe feed title left
- Swipe feed title right

Actions available:
- Save/Unsave story (taps only)
- Read/Unread story (taps, long press stories, swipe story titles left/right)
- Mark feed/folder read (long press feeds)
- Open original story (taps, long press story titles, swipe story titles left/right)
- Show original text (taps, long press story titles, swipe story titles left/right)
- Show menu (taps, long press stories/feeds, swipe stories/feeds left/right)
- Do nothing (all)

Just launched this in iOS, so I'd love to see them here. They are documented through the Preferences dialog which shows each gesture and it's multi-option select of actions.

PR #963, opened 2016-11-03
Android: Thumbnail Tweaks, Bugfixes, Beta Gestures

dosiecki:
Implemented feedback from last PR (#942) w.r.t. thumbnails.  Fixed a few related bugs from the first beta.  Fixed a few bugs reported by users.  Did the requested config/discoverability to get a very minimal implementation of gestures (#496) out to beta, by popular customer request.

Ready to go to Beta channel.


samuelclay:
Looking good. Few visual changes I'd like to see.




 * On iOS the unread icon is vertically centered with the title, not the row.
 * The author and date is bottom aligned in the row.
 * The content is vertically centered in the row, which means the title should have a minimum height of two rows.
 * The border on iOS is a lot lighter.
 * The date color on iOS is #262C6C.
 * The left-hand feed bar colors must be using the wrong attributes. Use 'favicon_fade' and 'favicon_color'.

samuelclay:
Also, users have mentioned that the long press menu shows up in the wrong spot in the beta: https://getsatisfaction.com/newsblur/topics/android-long-press-menu-appears-in-wrong-spot

dosiecki:
> On iOS the unread icon is vertically centered with the title, not the row.

This took some juggling (since the title was to the right of the dot, setting the dot to be centred to the height of the title would have created a cyclic layout dep), but figured out a way to rearrange the layout process to make it work.

> The author and date is bottom aligned in the row.

I reduced the bottom margin to be as thin as anywhere else, but this was probably ~1physpx on most devices.  If you're seeing the rare case where there is an image preview but not a text content preview and the author/date float up a bit, this will take a bit more work to solve, probably involving a custom viewgroup.  This is, oddly enough, bad enough to justify a separate task.

> The content is vertically centered in the row, which means the title should have a minimum height of two rows.

To my knowledge, there isn't a way to say that a TextView should have a height of exactly two rows and have it work when there aren't two rows of text, without some hackery.  We have it so that it will be at most two rows, and the other components will flow appropriately.  As with the above item, this might actually merit a whole task to track if desired.

> The border on iOS is a lot lighter.

Lightened it a bit more.  There are two global colours for the row-depth effect we've created.  The highlight was already FF white and the other is about as close as we can get without risking the lowlight being paradoxically lighter than the background it is ofseeting, due to the unreliability of the background colour in some tasks.

> The date color on iOS is #262C6C

I de-saturated this a few passes ago so it matched the web UI, figuring visual consistency between web/Android was more important than iOS/Android. (and because coloured metadata fields are, I believe, counterindicated by the design guidelines)

> The left-hand feed bar colors must be using the wrong attributes. Use 'favicon_fade' and 'favicon_color'

First time I am seeing this code, oddly!  Looks like we were using `favicon_border` and `favicon_color`?  Will swap out the former for the `_fade` variant, assuming I'm not just mixing up aliases.

samuelclay:
Sounds good to me. That commit looks like it does everything needed, is there more for the PR?

dosiecki:
Was spending some more time trying to track down a sync issue reported in GetSat, but I think I've done all towards it that I can for now.  Will do a PR unless #965 is a blocker.

PR #967, opened 2016-11-22
Android: bugfixes

dosiecki:
Fixes and improvements for the 5.0.0 beta, ready to become b2.


samuelclay:
Oh gosh, this is so, so beautiful. Releasing to the beta channel. I'd like to get this out before we put notifications in so we can have 2 android releases. You feeling ready for a release?

dosiecki:
Yep, should be good to go after another several daysa to soak in Beta.  The massive resource load for the new thumbnails has countless edge case effects, so the time in Beta is proving quite fruitful, and I'd like to ensure we actually caught everything in this patch.

PR #968, opened 2016-11-22
Android: Support for muting feeds

manderson23:
Finally got time to finish this off. Screenshot shows how muted feeds appear in all view.

screenshot_20161122-230240


samuelclay:
This looks good. We don't show muted feeds in the All mode on either web or iOS, but I think it might be nice to try it out on Android.

clients/android/NewsBlur/src/com/newsblur/util/FeedUtils.java
```diff
@@ -270,6 +292,38 @@ protected void onPostExecute(NewsBlurResponse result) {
...
+
+    public static void unmuteFeeds(final Context context, final List<String> feedIds) {
+        updateFeedActiveState(context, feedIds, true);
+    }
+
+    private static void updateFeedActiveState(final Context context, final List<String> feedIds, final boolean active) {
+        new AsyncTask<Void, Void, NewsBlurResponse>() {
+            @Override
+            protected NewsBlurResponse doInBackground(Void... arg) {
+                APIManager apiManager = new APIManager(context);
```
dosiecki: Generally speaking, we have been trying to contain non-readonly API actions to being performed by the SyncService.  This allows them to be executed in a predictable order, with correct retry logic, and without creating race conditions.  The framework for this is the ReadingAction class.  Doing this API call in an AsyncTask will probably cause crashes or data loss if gets called in an offline or network brownout state.

manderson23: OK. I'll take a look at changing it.

PR #981, opened 2017-01-06
Android automatic refactor - Recycle

luiscruz:
Hi,

I am developing a tool to automatically refactor Android applications with the goal of improving energy efficiency.
This pull request has the changes generated while applying the rule "Recycle".

Some resources (e.g., ```Cursor``` instances) should be closed when they are no longer necessary. 

I have made a previous validation of the changes and they seem correct.
Please consider them and let me know if you agree with them.

Best,
Luis


dosiecki:
Your tool is mistaken.  The cursor is very clearly closed.

luiscruz:
You're right!!
It is a common case that is not supported yet.
Sorry for having made a PR for this!

PR #982, opened 2017-01-06
Android automatic refactor - ObsoleteLayoutParam

luiscruz:
Hi,

I am developing a tool to automatically refactor Android applications with the goal of improving energy efficiency.
This pull request has the changes generated while applying the rule "ObsoleteLayoutParam".

While developing your application's views you might be specifying attributes in a view's artefact that are not necessary due to the nature of its parent. In this PR, those attributes were replaced by a comment.

I have made a previous validation of the changes and they seem correct.
Unfortunately, this tool is not able keep the original whitespace of the files, so comparison without ignoring whitespace might be confusing.
Please consider the changes and let me know if you agree with them.

Best,
Luis

### DIFF
--- a/clients/android/NewsBlur/res/layout/row_feed.xml
+++ b/clients/android/NewsBlur/res/layout/row_feed.xml
@@ -1,101 +1,93 @@
-<?xml version="1.0" encoding="utf-8"?>
-<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<?xml version='1.0' encoding='utf-8'?>
+  <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+  android:layout_width="fill_parent"
+  android:layout_height="wrap_content"
+  style="?selectorFeedBackground">
+
+  <LinearLayout android:id="@+id/row_feedcounters"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:layout_alignParentRight="true"
+    android:layout_centerVertical="true"
+    android:layout_marginRight="9dp"
+    android:gravity="right"
+    android:orientation="horizontal">
+
+    <TextView android:id="@+id/row_feedneutral"
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:layout_marginRight="3dp"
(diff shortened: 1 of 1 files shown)


dosiecki:
You may wish to update your tool before running it on anyone else's repos, as it is again wrong in this case.

The tool make nearly 100 noop whitespace changes and removed 1 layout parameter that does impact rendering on some platform versions. (ListViews have a number of bugs regarding properly pre-measuring layout spacing on contained elements that are somewhat deeply nested. I believe the parameter in this case is one of them.  Run your changed code on a few older platforms and change the font sizing and you'll see what I am on about)

luiscruz:
Hi @dosiecki 

Thanks for taking a look into this and for your feedback.
The loop whitespace is indeed an issue.
I have to figure out a better way of manipulating XML files.

In this case, the ```layout_centerVertical``` attribute was removed because ```ImageView``` is inside a ```LinearLayout```.

I didn't know this was useful on older platforms.
Could you please tell me in what version of the SDK you have found it necessary?

Thanks and sorry for the noise!
Luis

dosiecki:
The general problem is with dynamically sized rows within the classic ListView widget.  RelativeLayouts with wrap_content sizing do not correctly size or align children with match_parent heights to fill the row if the height of the child views are dynamic (in this case, we have a double problem because some of the imageviews are dynamicaly loaded and because text sizing can change).  Wrapping some components in seemingly-unnecessary LinearLayouts helps in some cases (I haven't ever had the time to peruse the platform code to figure out why).  In others, alignment params that would seem to be pointless ensure the child views stay where they should when parent re-sizing does occur.

I don't have the bandwidth to go back and do any specific testing for this, but if you run our codebase on a test device (I think I last personaly saw this on a 4.x series build) and start removing seemingly pointless layout params, you will eventually hit cases where some child views go invisible.  Thus as a general rule, we don't touch any of the row layouts unless addressing a bug that is servere enough to justify fairly comprehensive testing.

In the future, please build and test your changes on at least a 4.x - 7.x devices and pay especial attention to row rendering.  I'm not even 100% certain you would hit this bug in all cases, but it is common enough that I don't want to touch anything here without having done the testing. (my apologies for not having a full Testing Procedure Specification for this app)

PR #983, opened 2017-01-09
Android: Remove mute/unmute options for social feeds

manderson23:
Matches web UI and fixes crash if either option was chosen with a social feed.


samuelclay:
Nice catch.

Issue #966, opened 2016-11-22
Android: Story notifications


samuelclay:
This is a big ticket feature. There are two parts to this feature, this ticket only focuses on the actual notification. Choosing notifications for feeds is the next ticket, but we can launch that later.

0) Go on the web and right-click feeds to setup "Notifications" for android. You should be able to see it as it's available to admins.

1) After `/reader/feeds?flat=true` is loaded, loop through every feed and check if 'android' is in any feed.notification_types. If so, request notification privileges on the device.

2) Once android gives you a notification token that can be used for pushing remote notifications, send that token to POST to `/notifications/android_token/` with `android_token` as the key. This is idempotent, so you can submit this over and over again with the same token. That's what we do on iOS, to handle changing tokens. This endpoint is also device-aware, so multiple devices can submit different tokens to the same endpoint and each device will receive notifications.

Use `http://www.newsblur.com/notifications/force_push?feed_id=X&count=N` to test notifications. Feel free to hit it as much as you like.

Ideally I'd like to get images in the notification preview, as I did on iOS.

![photo nov 21 5 15 53 pm](https://cloud.githubusercontent.com/assets/44229/20507281/90ab1602-b00e-11e6-8beb-84575d6943cb.png)
![photo nov 21 5 16 14 pm](https://cloud.githubusercontent.com/assets/44229/20507282/90c43402-b00e-11e6-8e2e-bc7d15753183.png)
![photo nov 21 5 16 17 pm](https://cloud.githubusercontent.com/assets/44229/20507283/90c4994c-b00e-11e6-979a-9742c65db11a.png)

samuelclay:
Mark read and Save story actions do not open the app, they just silently work. Tapping on the notification opens the feed and then finds the story.

dosiecki:
Is there any background documentation on what a token is and the design rationale for this?  The app already has all the metadata needed to know when to post a notification that a feed has a new story without the whole extra round-trip to the server.  That is, if `notification_types` has the android flag, can't the app just post a notification when those feeds have a new story?  

Also, from a system standpoint, one-notification-per-story will likely not work.  Individual notifications can be heavyweight to process on some devices and the system will limit them and/or break if they are overused.  The typical design pattern on Android for a UX like this is to have a single notification that says "you have N new stories" and it changes as the count does rather than stacking up notifications for each new story.

samuelclay:
[...]

PR #988, opened 2017-02-08
Android: Second Notifications Alpha

dosiecki:
As described in #966, another incremental pass at building Android support for notifications.

### DIFF
--- a/clients/android/NewsBlur/AndroidManifest.xml
+++ b/clients/android/NewsBlur/AndroidManifest.xml
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.newsblur"
     android:versionCode="133"
-    android:versionName="5.0.0" >
+    android:versionName="5.1.0b1" >
 
     <uses-sdk
         android:minSdkVersion="16"
--- a/clients/android/NewsBlur/src/com/newsblur/database/BlurDatabaseHelper.java
+++ b/clients/android/NewsBlur/src/com/newsblur/database/BlurDatabaseHelper.java
@@ -977,16 +977,12 @@ public class BlurDatabaseHelper {
         return c;
     }
 
-    public Cursor getNotifyStoriesCursor() {
-        return rawQuery(DatabaseConstants.NOTIFY_STORY_QUERY_BASE, null, null);
+    public Cursor getNotifyFocusStoriesCursor() {
+        return rawQuery(DatabaseConstants.NOTIFY_FOCUS_STORY_QUERY, null, null);
     }
 
-    public void markNotifications() {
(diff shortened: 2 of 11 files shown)


dosiecki:
Just a heads-up, I am already tracking down one rare-ish bug where if you tap into two stories from notifications too quickly, they might lag until you rotate your device.  Working on a workaround for the next round of fixes.

samuelclay:
Ok, it's looking good. I did get 5 notifications as I loaded the app, which is a weird experience as I was in the app already. I think that won't be a problem since the app updates itself in the background.

By the way, I keep having to turn on Download Stories. Can we ensure that this option is checked by default?

Also, what's the story with background updates. Is there a preference for this anywhere? On iOS, background updates are unreliable and infrequent.

dosiecki:
> which is a weird experience as I was in the app already.

Should notifications not pop if the app is in the foreground?  I could easily change this behaviour, but I figured it was still appropriate for them to show, since you might have thousands of stories on screen but still want to be able to jump to the notification-enabled ones with a single tap. 

> I keep having to turn on Download Stories. Can we ensure that this option is checked by default?

I would love to default this to enabled, totally unrelated to Notifications.  It makes the UX so much smoother and honestly has few ill side effects.  Most users seem to enable it, too, from what I have seen.

> Also, what's the story with background updates. Is there a preference for this anywhere?

As it turns out, background updates are the *only* way the app communicates with the servers; there is essentially no such thing as a foreground sync.  This is a pretty common Android design pattern.  The UI uses a pair of event busses to communicate with the background service to do real-time stuff.  If the user disables background story sync (or if the network is not of the authorized type), the background sync just stops a little bit earlier.

samuelclay:
Yeah, let's turn off notifications for stories that come in while in the app. Notifications are intended as a background notification.

dosiecki:
> Yeah, let's turn off notifications for stories that come in while in the app. Notifications are intended as a background notification.

Finally having found a few use cases for notifications and using them for a few days, this goes rather counter to their value for me; if I spend an hour reading in one feed, I absolutely want to be notified if I get a story of interest in a different feed.

Is it just the sudden pop of notifications that you got when first enabling them that is the problem, or am I an outlier that I want to see notifications even if I am buried elsewhere in the app?

samuelclay:
I think it's jarring to have notifications, which are meant for you to see when you are elsewhere on your phone (in another app) or rolled up when you return to your phone, pop up when you're initially loading your feeds.

samuelclay:
I'm fine with it though, I just wish the stories would come in as push notifications and not when you loaded the feeds.

I do have an issue with only 1 feed coming through...


screen shot 2017-02-17 at feb 17 1 45 21 pm

PR #1001, opened 2017-03-15
NewBlurViewController_ANF

TopCruiser:
finished on converting ASI calls into ANF calls.
but I realized that all classes in Offline directory has been implemented using ASI downloading queue, so we should convert into or replace them with somethinig new.


samuelclay:
I'm sorry Michel but I'm not going to be able to accept this PR. It's missing some of the subtleties that come from institutional knowledge, and later on there are trickier migrations that need more knowledge about the entire code base. It was my mistake to outsource this work.

Issue #989, opened 2017-02-08
Android: Action/Sync Queue Can Become Blocked


dosiecki:
Over the last several months we have had very small but very steady number of bug reports from users about the app refusing to sync.  Debug info received from them shows a comon cause of the action queue becoming poisoned by blocking requests.

More recently, the sync logic was changed so that a *failed* action in the queue would not block the queue or the sync loop, on the suspicion that it was merely fatally bad actions poisoning the queue.  This improvement has not eradicated the issue, suggesting that the blocking actions are actually blocking in a network/execution sense and not just failing and being retried.

The commonly effective workaround of clearing app data suggest that there may exist a class of actions that are fatally impossible to incomplete.

Ideally, it would be great if we could devise a mechanism for getting raw debug/log info from users that experience this bug, so we can fix the root cause(s), since, whatever they are, they have totally evaded reproduction.  In the meantime, it may be necessary to break the in-order call contract and risk slight data loss about reading actions if the problem gets any more common.

dosiecki:
Collecting for posterity some debug reports we have received that may be related to the issue:

* https://getsatisfaction.com/newsblur/topics/android-bemnmpwir2mmw
* https://getsatisfaction.com/newsblur/topics/android-app-not-working-rwkshjp896mbd
* https://getsatisfaction.com/newsblur/topics/android-new-version-no-unread-stories
* https://getsatisfaction.com/newsblur/topics/android-delays-reading-any-already-fetched-articles-while-catching-up-on-reading-actions-for-tens-of-seconds
* https://getsatisfaction.com/newsblur/topics/android-catching-up-reading-actions-forever
https://getsatisfaction.com/newsblur/topics/android-beta-5-0-app-could-not-find-any-new-stories-had-to-uninstall-and-reinstall

dosiecki:
A couple of theories to pursue if this gets bad enough to do a deeper dive:

1) DB state gets corrupted or mismatched to structure in some way to create un-runnable actions. No direct evidence, but the fact that the bug sometimes happens right after upgrade points us in this direction as it is the one time the DB structure changes.

2) Some internal lock, file, or resource is getting raced for.  This hypothesis is a result of the observation that these reports just happened to start around the time story list thumbnails were launched, which use a great deal of resources asynchronously to load.

[...]

PR #1006, opened 2017-03-31
Android: Full Notification Support, Bugfixes

dosiecki:
* update build system to explicity fail fast on insufficient java version
* don't unnecessarily reset API pagination
* cancel outdated / stacking DB requests
* explainer for zero stories due to Saved mode
* more diagnostics targeting #989 
* fix notification taps bringing up wrong story
* make notification dismissals sticky
* notification actions
* preference to enable notifications (no longer gated behind unrelated option!)


samuelclay:
Pushed out as 5.2.0b1. Also released v5.1.0 to prod just now.

dosiecki:
Excellent.  One thing in particular that needs beta testing is the behaviour of tapping directly into notifications.  Experience suggests we'll get some slightly odd edge cases on devices with custom OEM roms when launching the activity repeatedly.  We should encourage folks to send along any logs using the new feedback mechanism if they see anything at all odd!

samuelclay:
Hey, nicely done. In my limited 24 hours of testing it's working great. I'm giving it the full test over the next 24 hours, syncing it up with what I have on my phone. And I really like how fast everything works. 

Next we should think about getting the notification editor in there. I think you have access on the web. Here's what it looks like on iOS. I also made swiping right a shortcut for notifications when swiping feed titles.

img_0923
img_0929

PR #1011, opened 2017-04-20
Android: Custom Story Fonts

manderson23:
Support for custom fonts in stories. The 3 OTF files and Noto are available.

@dosiecki mentioned via e-mail he would like a condensed and monospaced web font as an option. How about [Droid Sans Mono](https://fonts.google.com/specimen/Droid+Sans+Mono) and [Open Sans Condensed](https://fonts.google.com/specimen/Open+Sans+Condensed)

Also I wasn't sure about "Default" as the option for just using the platform font. I'm open to suggestions.

I've only tested on 7.1 so wider testing is definitely required.


dosiecki:
Can we load Chronicle / Gotham / Whitney from the web, too?  They must already be hosted on the API domain if they are available on the web, no?

As for additional fonts, I would say both Noto Sans and Noto Serif are the must-haves.  Open Sans Condensed looks pretty good for a condensed.  I think I slightly prefer Inconsolata or Anonymous Pro for monospace.

samuelclay:
I believe you'll need to embed them directly. They are already in the repo somewhere.

samuelclay:
Keep those fonts on the device itself. We load them on iOS as well. It doesn't add much and we don't need to worry until the app is more than only a few MBs big.

manderson23:
This should hopefully be good to go now.

samuelclay:
This looks awesome. Thanks so much for putting this together!

Now, can we get a font menu on the story detail view? It'd be nice to be able to switch between fonts and compare. Perhaps similar to how we do font size switching.

manderson23:
No problem I'll add that.

samuelclay:
I'm writing up the latest Android release. Here's what I've got so far:

 * Mark as read on scroll is a new preference to automatically mark stories as read as you scroll past them.
 * New fonts for reading: Whitney, Chronicle, Gotham Narrow make their way to Android from the web.
 * Switching between Text and Story view is now sticky per-feed (similar to iOS and web).
 * Fixed issues where feed list would not update.

Anything to add that I'm missing?

manderson23:
The additional Android fonts - Noto Sans, Noto Serif, Open Sans Condensed and Anonymous Pro.

dosiecki:
> I'm writing up the latest Android release. Here's what I've got so far:
> Anything to add that I'm missing?

In terms of what's out to Prod, the big one would be notifications.  However, I would hold off at least another day.  I've got 3-4 fixes in flight that need to go out. (almost nobody appears to be using notifications in Beta, so we're getting no test feedback)

PR #1016, opened 2017-05-04
Android: Font option in story menu

manderson23:
As requested added a font option to the story menu


samuelclay:
This is great stuff. Can you make the font choice dialog sticky after choosing a font? Similar to how the font size dialog is sticky until you tap in the background to dismiss.

PR #1020, opened 2017-05-31
Android: Mark-on-Scroll, Beta Bugfixes

dosiecki:
Implements the new mark-on-scrol option and fixes several bugs present in prod and beta.  Also did a pass of cleanup on warnings/lint/style.

Note that I did finally give in and bump our minimum platform version from 4.1 to 4.4 in line with what GOOG have stopped supporting.


samuelclay:
Interesting, you chose to mark the story as read as soon as the top of it gets to the top line, not the bottom of the story (as we do on iOS and web). In other words, on android you see stories being marked as read as they scroll. On iOS and web, you don't. I think I may change it the Android method.

samuelclay:
Release as 5.2.0b4.

dosiecki:
> Interesting, you chose to mark the story as read as soon as the top of it gets to the top line, not the bottom of the story (as we do on iOS and web). In other words, on android you see stories being marked as read as they scroll. 

Yeah, pinged a few threads about the feature for tips and didn't get much of a response, so went with my gut.  I figured that if people generally read top-down, they are "read" as soon as they are seen?  I couldn't get the mark-on-scroll option to work on the web, so I went with that.

If it turns out to not be a great approach, the other impls. I could swap to without too much time cost would be to mark when a story (assuming top-down reading withing the list):
* fully passed out of view to the top of the screen
* hits a midpoint in the list
* shows up in part at the bottom of the screen

samuelclay:
No, I love it this way. I changed iOS to mark as read as soon as half of the title of the top story goes above the viewport.

samuelclay:
Let's turn this on by default, as we're going to be doing on iOS and web.

PR #1023, opened 2017-06-07
Android: Beta Bugfixes

dosiecki:
Fixing a few bugs found in the latest beta and a few found by keen users on the forum.  Feature freezing from here forth so we can, with luck, take this one to Prod.

### DIFF
--- a/clients/android/NewsBlur/res/values/strings.xml
+++ b/clients/android/NewsBlur/res/values/strings.xml
@@ -145,6 +145,7 @@
     <string name="toast_story_unread">Story marked as unread</string>
 
     <string name="logout_warning">Are you sure you want to log out?</string>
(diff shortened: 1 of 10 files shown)


samuelclay:
Just got the iOS app approved, so I'm ready to launch whenever you are.

samuelclay:
Released as 5.2.0b4.

dosiecki:
Are any beta users actually unlocked to use notifications other than we three devs?  I've had nothing in the way of feedback or crashes from the notification subsystem, and I would be curious to know if that is a false negative due to a lack of usage!

samuelclay:
I haven't promoted it and because we're not sending Android notifications from the server, I can't track it either. I figure we'll fix any bugs as they come in, so long as the base system is there and works. Plus Android's quicker to fix than iOS.

dosiecki:
Small set of fixes incoming.  Some of these are imperfect, but they are the best we can do with the limited testing data we've got so far.

PR #1025, opened 2017-06-12
Android: Bugfixes

dosiecki:
Small fixes to the last beta.  Most of these were so rare to repro, it is probably best to just get them to Prod and iterate as quickly as possible therein.

### DIFF
--- a/clients/android/NewsBlur/src/com/newsblur/activity/ItemsList.java
+++ b/clients/android/NewsBlur/src/com/newsblur/activity/ItemsList.java
@@ -168,6 +168,7 @@ public abstract class ItemsList extends NbActivity implements StoryOrderChangedL
                 searchQueryInput.requestFocus();
             } else {
                 searchQueryInput.setVisibility(View.GONE);
+                checkSearchQuery();
             }
         }
 	
--- a/clients/android/NewsBlur/src/com/newsblur/database/StoryItemsAdapter.java
+++ b/clients/android/NewsBlur/src/com/newsblur/database/StoryItemsAdapter.java
@@ -81,7 +81,7 @@ public class StoryItemsAdapter extends SimpleCursorAdapter {
 		return cursor.getCount();
 	}
 
-    public boolean isStale() {
+    public synchronized boolean isStale() {
         return cursor.isClosed();
     }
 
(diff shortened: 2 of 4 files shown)


samuelclay:
Something feels different about this build. The story list UI now redraws a bunch and after it renders an image the image then disappears and then reappears, all while redrawing the list and bumping everything up and down. I see those `synchronized` keywords being added. In Obj-C, doing that leads to a performance hit as it holds a lock on reading any instance variables. Could that be happening here?

samuelclay:
There seems to be a slight pause when a story is marked as read on scroll. Can we figure out how to scrub that pause? Is the mark as read action in a background thread? Can we queue it up and then batch it?

dosiecki:
> There seems to be a slight pause when a story is marked as read on scroll. Can we figure out how to scrub that pause? Is the mark as read action in a background thread? Can we queue it up and then batch it?

Yep, it is all choreographed in the background, like all API actions.  This isn't a feature I ever use, so my testing has been a bit limited, but on my 5X I get almost perfectly smooth scrolling with it turned on.  Can you do a debug/log capture to me on whatever device you are seeing pauses so I can look into it?

samuelclay:
I guess I didn't let it settle enough, because now it seems to be smoother. Ok, I'll launch it tomorrow then.

dosiecki:
I'll definitely say - if actively using the UI with even some of story prefetch, notifications, thumbnails, and quick-mark enabled, the app puts a very heavy load on all but the best and newest devices.  Users on slightly older or more resource-constrained devices may not be able to use all of the new features simultaniously and still get a silky-smooth experience, unless we were to devote a fair bit of time to optimizing or replacing a few of the stock components on which we've built.

dosiecki:
> Something feels different about this build. The story list UI now redraws a bunch and after it renders an image the image then disappears and then reappears, all while redrawing the list and bumping everything up and down.

If this is happening just after enabling mark-on-scroll, I'm not entirely surprised.  Those list views aren't really designed for in-place modification.  Previously, items in the story list rarely changed state when visible (most state changed happened in the Reading view or when marking-all-read and switching screens), but the new read-state manipulation bucks that trend.  The more features we add to the story list (thumbnails, dynamic state, etc) the more we push past the design boundaries of the stock list widget.  There is a new framework as of Android 6 that allows us to write custom lists widgets with a much greater speed, should we want to invest that time.

> I see those synchronized keywords being added. In Obj-C, doing that leads to a performance hit as it holds a lock on reading any instance variables. Could that be happening here?

That is part of a compatibility fix for some changes made in Android 7/8 to how background data are loaded into the foreground UI.  It shouldn't cause any noticible pausing except for in cases where the app would previously have crashed (so, comparatively rarely).

PR #1027, opened 2017-06-22
Android: 6.0.0 Bugfixes

dosiecki:
High-priority fixes from the 6.0.0 staged rollout.  Would recommend quick internal testing and pushing to a 6.0.1 staged rollout before 6.0.0 gets any further.


samuelclay:
Looks good to me. I did experience one issue: while browsing by folder I shared a story and while it was shared, the story never updated to show the share.

samuelclay:
Ok, rolling out 6.0.1 to 20% of prod. I didn't realize I had to manually up the %age. No wonder a bunch of users emailed me about v5.1.0 bugs that I figured they'd have fixed by now. I'm going to roll this out to 100% if we don't have any show stopper bugs on Monday. I wish roll out would just do 10% and then 90% automatically in 24 hours.

dosiecki:
> while browsing by folder I shared a story and while it was shared, the story never updated to show the share.

Noted.  Definitely not a regression; there are a few edge cases around social metadata where we don't accurately update local state to match remote state until a deeper sync.

Issue #177, opened 2013-04-18
Android: Keep read sites visible until refresh


samuelclay:
Site has 1 unread, go in, read it, come back to feed list. Should still see the feed. It should disappear on refresh.

dosiecki:
Implemented in #1028 / #1033

PR #1028, opened 2017-07-05
Android:  Bugfixes

dosiecki:
Fixing some uncommon crashes seen in Prod.  Also fixing some long-standing bugs (such as #177).

Disables some social features that appeared semi-working in Prod but were fundamentally flawed and causing user confusion.  Added TODOs where these need fixing.  These changes open up the addition of some new social features (reply and comment editing) that were previously blocked.

This should probably *not* be considered ready for release, since it in some ways will appear as a superficial regression to some users.  There is some work yet to be done with regard to the use of the social API.


dosiecki:
Just to elaborate for anyone curious about the intentional regression of the social features:

The API generates UIDs for all social shares / comments / replies that are required to deduplicate or manipulate them.  The existing code faked these IDs so it could perform social operations locally without delay.  This, however, led to some weird UI duplication and/or lag, and also prevented us from adding missing functionality around share/comment/reply modification.

I've temporarily disabled the hackish local implementation of social identifiers, since it was fundamentally broken.  As a resultk this build is *not* ready for prod.  It still correctly sends social actions to the server, but relies on a full sync loop to reflect them in the UI.

The next PR will attempt to reconcile the server-assigned social UIDs with locally created data based upon immediate API calls to restore the expected UX in a slower but more reliable way.

samuelclay:
Why bother with the local ID? On iOS, I just freeze the UI while sharing is being sent to the server and redraw when it comes back. And if it fails, the comment text is still in the dialog, ready for editing and copying and pasting for later.

dosiecki:
> Why bother with the local ID?

That is what I removed.  Relying on a remote ID will slow down the UI by a lot, but that seems unavoidable unless the ID can be made into some sort of predictable hash of some sort.

Issue #1030, opened 2017-07-22
Web: Text wraps instead of converting to ellipses in Edge


ctafield:
On Edge, titles of feeds wrap around, instead of truncating to ellipses (like in Chrome etc...)
![image](https://user-images.githubusercontent.com/2324109/28490606-2645ffde-6ed6-11e7-9b2e-3741ece43a81.png)

I believe this is down to a missing style on .NB-feedlist .feed_title. Looks like it needs to have:
`white-space: nowrap;`

and perhaps the display changing to block... (for Edge).... Not sure (sorry, can't repo this properly locally on my setup).

Adding to it. That seems to fix it in my brief tests, like such:
![image](https://user-images.githubusercontent.com/2324109/28490613-536238ca-6ed6-11e7-8425-828df84adce9.png)

Apologies for not being able to offer up a pull request!

samuelclay:
You fixed this just fine in #1031. I'd love to merge it, just submit the request.

PR #1031, opened 2017-07-22
Fix for Site titles in Edge

ctafield:
Fix so that titles are truncated with an ellipsis in Edge browser. https://github.com/samuelclay/NewsBlur/issues/1030

Before:
image

After:
image


samuelclay:
Hey, what happened to this fix? This is perfect and I would have accepted it. If you re-open it I will gladly merge and deploy it.

Issue #1032, opened 2017-07-27
Android: insta-fetch stories menu item


samuelclay:
Stick it in the Feed Detail menu. It should POST to /rss_feeds/exception_retry with the `feed_id`. Only available for individual feeds.

dosiecki:
Implemented in #1033.

PR #1033, opened 2017-08-02
Android: Social Fixes, Insta-Fetch, Bugfixes

dosiecki:
* Proper fixes for the issues discovered with social features last month.  All social features should now be working as intended with a snappy local UX, full async/offline support, and immediate/intermediate/final state that matches the web UI.  Also added the missing functions for deleting and editing comments and replies.
* Fixed the broken no-focus-unreads explainer; added some logging to help us quickly diagnose user confusion about focus mode.
* Improved handling of stories with failed original text extraction. Now messages user and switches them back to a mode with content to view.
* Implemented insta-fetch for feeds. (#1032)
* Various non-critical bug fixes.

Should be ready for a week in Beta and then a week in 10% rollout, at least.

### DIFF
--- a/clients/android/NewsBlur/AndroidManifest.xml
+++ b/clients/android/NewsBlur/AndroidManifest.xml
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.newsblur"
     android:versionCode="141"
-    android:versionName="6.0.0" >
+    android:versionName="6.0.2" >
 
     <uses-sdk
         android:minSdkVersion="19"
--- a/clients/android/NewsBlur/res/layout/row_feed.xml
+++ b/clients/android/NewsBlur/res/layout/row_feed.xml
@@ -72,6 +72,14 @@
             style="?muteicon"
             android:background="@drawable/neutral_count_rect"/>
 
+        <ProgressBar
+            android:id="@+id/row_feedfetching"
+            android:layout_width="21dp"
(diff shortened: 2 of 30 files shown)


samuelclay:
Sweet, just pushed out as 6.1.0b1 to 507 beta testers.

manderson23:
I've noticed one issue with the latest changes.

If I'm in unread only view, I navigate into a feed and mark all read.

When it goes back to the main unread only view the row for the feed I've just marked as read is still visible but with no number next to it. 

When I move into another feed/folder then go back it disappears as expected.

samuelclay:
I love this behavior, it's how I have it on the web and ios. I believe I asked for it then. You should see the last feed you were on, ideally with an animated fade out. iOS does this for free, so don't bother building it unless it's also free, but it would be nice if the highlight color faded out upon return. Same with story titles.

samuelclay:
Also, a user agrees: https://twitter.com/flameeyes/status/895589834533818368

screen shot 2017-08-10 at aug 10 12 40 50 pm

manderson23:
Ah OK. It doesn't fade on Android but just sticks around until the next action that triggers a UI update.

dosiecki:
> When it goes back to the main unread only view the row for the feed I've just marked as read is still visible but with no number next to it.

Yep, this was requested a very long time ago in #177.  I'd been waiting for a chance to implement it with the correct animation/fade, but upon the realization that that is almost impossible without some massive rewriting, I put in the more basic version.

PR #1036, opened 2017-08-11
Firefox Fix: change host to match Firefox iOS app change

joeyg:
There was a change to the Firefox iOS app where linking now requires a host of "open-url". This change broke linking to firefox from newsblur and this PR addresses the issue.

The corresponding code in Firefox is here: https://github.com/mozilla-mobile/firefox-ios/blob/master/Client/Application/AppDelegate.swift#L393


samuelclay:
Sweet, this will be in the next release. Although I don't have anything else going out in an iOS release, so it may take a little bit.

joeyg:
Thanks!

PR #1038, opened 2017-08-16
Support HTTP 308 from RFC 7238

da2x:


samuelclay:
I'm going on vacation but I'll look at this in September. I wonder if there are any other '301's that have to be extended?

da2x:
> I'm going on vacation […]

Then you should prioritize that before this nonsense. 🏝️

> any other '301's that have to be extended?

It’s only 301 and 308 that are permanent redirects. See [Redirect status codes and characteristics](https://en.wikipedia.org/wiki/URL_redirection#Redirect_status_codes_and_characteristics).

This may need to be updated as well, though I’m not sure if `HTTPRedirectHandler` handles 308 at all so that may need to be extended.
https://github.com/samuelclay/NewsBlur/blob/55c36a24f38c25561e64b4365767547e3945ac1b/utils/feedparser.py#L2766-L2778

This patch address the only place in Newsblur’s own code I could find that needed to be updated.

samuelclay:
Thanks for the PR! It's now deployed.

PR #1039, opened 2017-08-17
favicon fetcher: provide an Accept header on requests.

Flameeyes:
This makes the fetcher work with the default ModSecurity ruleset, and would also avoid providing unsupported favicon formats (e.g. webp), by explictly preferring PNG and
x-icon formats.


samuelclay:
Thanks! I've now deployed it so it may take a few days to refresh icons. LMK when it does, I'd love to know that this worked.

Issue #1029, opened 2017-07-19
Android: Subscribe intent


samuelclay:
We should build a share intent so that folks can subscribe to sites right from the site (or shared URL). I imagine 95% of this feature will be used from chrome. 

If possible let's do a v1 and v2. V2 is the in app dialog with async requests and confirmation. V1 just kicks the user over to the app or something relatively quick. More of quick way to get a URL subscribed and into a folder. We may just have to build v2. 

This has been a popular request for as long as I can remember and I think it's time we made it easier to get urls into newsblur.

dosiecki:
released in v6.1

PR #1040, opened 2017-08-30
Android: New Subscribe Intent, Reading Mode Rework, Bugfixes

dosiecki:
The Android app can now attempt to catch links to feeds an offer to subscribe to them. (#1029)  Implemented as broad a net as seemed reasonable.  Also made some improvements to the subscription UX so it is more obvious what is happening.

Reworked how view modes (Story/Original Text) work to fix some bugs introduced by the new handling of feeds that don't support Original Text extraction.  Also fixed the long-standing bug where a feed's view mode was not correctly used if stories from the feed were viewed in the context of a folder or a share.  Original Text fetching should also now be smoother and quicker.

Fixed some reported crashes.


samuelclay:
Release to beta as v6.1.0b2.

PR #1041, opened 2017-09-15
Android: Prod Release Candidate

dosiecki:
A few very small bugs caught in Beta.  App should be ready for immediate release to Prod.


samuelclay:
Nice, released as v6.1.0.

PR #1047, opened 2017-09-30
iOS: Fix duplicate title/URL in Messages sharing.

nriley:
Now Messages displays URL previews, just don’t include the title at all any more.

See https://getsatisfaction.com/newsblur/topics/app-share-sheet-duplicates-urls-shared.


samuelclay:
Thanks for this Nicholas! I've got some nice changes going for iOS 11 and should hopefully beta test for a week and push this out soon enough.

PR #1055, opened 2017-11-07
Android: Infrequent Site Stories, Bugfixes

dosiecki:
Support for the new Infrequent Site Stories feature, as well as a number of bugfixes for issues found in Prod.

Also includes some framework build-up for future improements to intel training from the app.


samuelclay:
Love the new infrequent site stories view. Nice job on the menu as well. The theme selector works like a charm. I just added the same theme selector menu to all views on iOS and it should be released at the same time.

I'm going to release this as beta tonight in the hopes that I can promote it to prod on Friday.

PR #1056, opened 2017-11-09
Docker: Use mongo:3.2 docker image in docker-compose.

hnrd:
Mongo 3.4 does not work properly with pymongo, so pin the mongo docker image to v3.2.


samuelclay:
Why doesn't mongo 3.4 work with pymongo? Should I just upgrade pymongo?

Issue #1058, opened 2017-11-18
Web: Long feed names has line breaks in new firefox quantum


creiner:
In the new firefox (Version 57.0) longer feednames are broken into mutiple rows, see picture:
![image](https://user-images.githubusercontent.com/137896/32979883-69581dea-cc5d-11e7-9cb4-1f7352d8ca6d.png)
Edit: Same in IE11

samuelclay:
Fixed in #1059.

PR #1059, opened 2017-11-18
Web: Changed whitespace break behavior in the feed title

creiner:
Fix for issue #1058. 
Needed to remove "display: -webkit-box", because "text-overflow: ellipsis" only works with "display: block".


samuelclay:
Sweet, thanks for the PR!

PR #1060, opened 2017-11-28
Get iOS 11 Smart Invert to leave images and videos alone.

mihaip:


samuelclay:
Excellent work as always. This will go in the upcoming v7.0, which I'm trying to get approved right now, but since I've submitted it I've already added in a few other iOS 11 specific features. So I'll probably resubmit later this week with this change.

Issue #1065, opened 2017-12-10
Android: Auto-open All Site Stories loads Infrequent Site Stories instead


samuelclay:
There's a preference for the app to automatically open a feed, right? Looks like it might be opening the wrong feed.

https://twitter.com/NewsBlurSupport/status/939940041559719936

dosiecki:
With the auto-open option turned on, tapped the All Stories row and got taken to the All Stories reading view.  Tapped Infrequent and got taken to infrequent.  Also did a code dive and manually traced out both paths and confirmed they have handling for auto-open, not some default behaviour.

I think maybe we're not understanding what OP means by opening "by default"?

samuelclay:
Is it possible it's the default scroll position?

dosiecki:
The folder/feed list always default to the very top - Global Shared.

samuelclay:
Yeah, based on the feedback i'm getting I'm going to probably request that we add three toggle settings: Show/hide global, show/hide all, show/hide infrequent. All Shared Stories shouldn't be hidable as that's a choice you make by following users.

dosiecki:
This seems like a pretty hot issue, so I went ahead an implemented options to disable either Global Shared or Infrequent Stories.  They both still show by default but have toggles in the Prefs panel.

I held off on the ability to disable the All Stories folder, since, due to some implementation details, if both All Stories and Infrequent Stories were disabled, it would make it so that top-level feeds had no parent header row to sit under, and would disappear.

dosiecki:
resolved in #1066

PR #1066, opened 2017-12-12
Android: Make Some Meta-Folders Optional

dosiecki:
As discussed in #1065, a quick patch to make optional some meta-folders that have been causing a bad UX for some folks.


samuelclay:
Can you also move Infrequent Site Stories to go immediately above All Site Stories?

PR #1067, opened 2017-12-16
swap infrequent and all rows

dosiecki:
as requested in #1066.

(sorry for brevity, made change from my phone)


dosiecki:
missed a block, will reopen

dosiecki:
fixed!

PR #1068, opened 2017-12-19
fix crash for users with no folders

dosiecki:
fixing the current bug in Prod


samuelclay:
Was this a remnant of having Infrequent below All and having the top-level feeds in Infrequent but act as if they're in All?

samuelclay:
I had to do some silly workarounds to do that, too, on iOS.

dosiecki:
It was something I missed when I swapped the position of the two rows at the last moment on Friday.  Pretty much all of our day-to-day testing happens on accounts that use folders.

PR #1069, opened 2017-12-26
Add "reply-to-self" tags to tweets

thinkpoop:
Add "reply-to-self" tags to tweets where the author and "in reply to" user have the same id


samuelclay:
Looks good, thanks!

PR #1071, opened 2018-01-04
Android: Black Theme, Caching Tweaks, Bugfixes

dosiecki:
Implemented the requested power-saving black theme.  Also tweaked some prefetch logic to save even more mobile data for users with feeds in original text mode.  Finally, fixed the one new crash seen in the last few Prod releases.


samuelclay:
Rolling out as v7.0.3. Black theme looks good.

PR #1073, opened 2018-01-06
iOS drag and drop improvements

nriley:


samuelclay:
Sweet, thanks so much!

PR #1077, opened 2018-01-28
iOS: Replace “magic number” 0-3 section indices with constants

nriley:
These indices are used in appDelegate.dictFoldersArray and as UIButton tags.

Removes a number of places where a number-to-name lookup was done for no apparent reason when the items are in appDelegate.dictFoldersArray already.

Fixes issue where all stories are now at index 3, not 2, and shift-E was showing infrequent site stories on iPad because the index was not updated when infrequent site stories were introduced.

### DIFF
--- a/clients/ios/Classes/FolderTitleView.m
+++ b/clients/ios/Classes/FolderTitleView.m
@@ -37,14 +37,7 @@
         [subview removeFromSuperview];
     }
     
-    NSString *folderName;
-    if (section == 0) {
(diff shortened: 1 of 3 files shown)


nriley:
(Sorry about that accidental review request — misclick.)

samuelclay:
This is great stuff, thanks for putting it together. I always felt it was brittle and it built up slowly over time, but I'm glad that this new solution makes it clear what's what.

PR #1078, opened 2018-01-28
iOS: Eliminate duplicate URL when sharing to Tweetbot.

nriley:
See e.g. https://twitter.com/acdha/status/956520307623383045.


samuelclay:
Nice work!

Issue #1076, opened 2018-01-18
Android: Code doesn't always scroll horizontally


samuelclay:
See https://newsblur.com/site/3667504/finding-the-missing-link

Scrolls on web and iOS but not Android.

<img width="739" alt="screen shot 2018-01-18 at jan 18 3 38 34 pm" src="https://user-images.githubusercontent.com/44229/35127169-a7730c24-fc65-11e7-8ada-d1bb46037be6.png">

![2018-01-18 15 37 12](https://user-images.githubusercontent.com/44229/35127195-c974d87a-fc65-11e7-8d4f-028b0b2519e2.png)

![2018-01-18 15 37 19](https://user-images.githubusercontent.com/44229/35127171-ac0611dc-fc65-11e7-9adc-53bc70fc501a.png)

dosiecki:
The poor horizontal scrolling cababilities of the in-built browser component are generally incompatible with a left-to-right scrolling pager.  Horizontal scrolling is explicity disabled for this reason.  We have gone back and forth over the years on whether or not code should be wrapped.  The overall recommendation is that to view complex content such as code, the story should be opened in a real browser.  Short of implementing our own web browser or getting rid of left-right paging, there is not a whole lot we can do that won't break browsing on a tangible number of devices.

This has been brought up so many times that I am starting to wonder if the app needs an FAQ.

dosiecki:
I should add that this is a great example of where Android's new open-in-app browser functionality can be used to good effect.  On moder devices, just tap the title of the story and it should open in a fully-functioning browser tab without even leaving the app where code should display just like in the full browser.  Getting back to the app is as quick at tapping the close tab button at the top or hitting the 'back' button.

samuelclay:
What happens when you style pre/code with overflow-x: scroll? It breaks paging for the entire page or just where there's code?

dosiecki:
We have horizontal scrolling totally disabled at the engine level.  The side effects of enabling it are myriad and ugly, due in large part to the fact that many devices' browser impls don't respect mad-width, width, or overflow properties, resulting in many stories (not just those with wide text blocks, but with any kind of large images or embeds) trying to scroll horizontally, blocking the left-right swipe action entirely, or causing bad screen flicker as the two scrolling portals fight to handle the drag events.

dosiecki:
[...]

Issue #1052, opened 2017-10-24
Android: Grid View


dosiecki:
Android is missing the grid-style story list mode that prioritizes thumbnails.  We should modernize the ItemsList activity to use a RecyclerView that will allow us to support more novel layouts than the vertical lists of old.

dosiecki:
Completed in #1095.

bingoxo:
can this be ported to ios client ? (please)

samuelclay:
Yep, this is now #1270.

PR #1080, opened 2018-02-02
Android: Bugfixes, Cache Control, Refactoring for Story Grid

dosiecki:
* fixed some market-reported crashes
* tweaked rendering of pre blocks to mitigate UX of wide block on small screens (#1076)
* added preference to let users control caching behaviour in case their devices don't respect the platform APIs w.r.t. cached files vs. permanent files and storage usage grows out of control (see https://getsatisfaction.com/newsblur/topics/control-storage-limit-in-android-app )
* totally refactored the story list fragment in anticipation of supporting a grid-style story list (#1052). decoupled the view layer from loading and startup logic.  the grid view will need to use a totally different view framework (RecyclerView) that is not backwards-compatible with the old ListView and Adapter paradigms.  we are now free to have many different story displays without having to rewrite loading and launch logic.

Due to the massive refactor required to support GOOG's new framework, this should soak in Beta for a week or two, as the change surface is massive and we should eliminate regressions in our loading logic before enabling the new framework code, which is reported to be somewhat buggy, though necessary.


samuelclay:
Deployed as v7.1.0 to beta channel.

PR #1085, opened 2018-02-23
Android: Grid View Prototype

dosiecki:
First pass at the grid view for the Android app is finally here.  All basic functionality should be working, though we'll need to make a pass of bugfixes and optimizations, for sure.  Also still need to add in some of the more advanced features such as gestures and column adjustments. (to come in next PR)

*First and importantly*: in order to add the dependency for the new framework that supports the grid, another set of Android framework libs had to be updated.  We had been previously been holding these back because GOOG forced some UI changes in them, so there will be some minor side effects outside of the grid view. (such as the new style pull-to-refresh indicaors on the feed list)  You may also need to double-check that your development environment is correctly picking up or importing dependency settings from the gradle config file to avoid build problems.

Performance is getting there.  This patch includes a significant rewrite of the way we load thumbnail images for stories.  The old loader would pretty always eventually OOM any non-flagship device, because some feeds specify thumbnails that can be many megapixels per story.  I added a system to do lossless streaming downsampling of thumbnails and it has helped quite a bit.  We'll probably need to go even a bit further, though, as the grid view still eats a good deal of RAM and CPU.

I tried to match the grid tile UI to the web UI, while still making good use of very limited screen space and respecting the Android UX guidelines.  Let me know if anything is far off the mark.

This should be released as an alpha, at most.  We'll need to do a fair amount of testing for regressions in both story list views, as many points were touched. (that said, some of this work also drastically speeds up the old list view when thumbs are enabled!)


dosiecki:
I'll also add that the new thumbnail loader is so quick that it unmasked some bugs in how the old list-style story list reloaded things.  Working to bring it up to speed ASAP.  Definitely an alpha release at most.

samuelclay:
Sorry I sat on this for so long. I just finished shipping Turn Touch and once I launch I'll have time to work on software for both NewsBlur and Turn Touch.

As for the grid, it's looking good. Can we add a number of columns option, like on the web? I find 2 columns to be too narrow for portrait and too wide for landscape. Maybe I can adjust the defaults somewhere? Also, the height in landscape mode is too tall. Is there a convenient place I can adjust these values?

dosiecki:
> Can we add a number of columns option, like on the web?

Yep, this is in the works (you'll find a very early version in this pull, but the real one will be adaptive to account for the fact that screens can be anywhere between 3cm and 3cm wide).

> the height in landscape mode is too tall

What do you mean by this?  The Height of the tiles is currently a function of the metadata displayed below them.  The only way to reduce their height is to eliminate or resize that metadata.

PR #1095, opened 2018-03-29
Android: New Story List Pt. 2, Bugfixes

dosiecki:
*(Note: this update includes substantial changes to dependencies; it is highly suggested to update to the latest version of gradle and to do a clean build)*

This is the completion of the story list rework.  The oldschool list view and the new grid view have been merged to both use the newest recommended framework.  As a result, all styles of story list now support all options (such as gestures) from a single, cleaner codebase.  All story lists also run at about 2x their old framerate, even with thumbnails enabled!

One big component of the new list impl involved upgrading virtually every support library dependency and dealing with 3+ years worth of breaking changes therein.  The chain of dependencies forced upgrades in basically every framework we use.  Chased down as many regressions as I could and also used some newly available features to resolve some old minor bugs.  (Notably, some of the more annoying story Pager bugs have been reduced or eliminated!)

In order to pull up said dependencies, we finally have to drop support for devices running Android 4.4 - 4.4.4, making 5.0 the new minimum, as the pre 5.0 builds are now totally unsupported by GOOG.

Also took care of a few reported crashes and bugs from the live release.

Ready for Beta.  In spite of dozens of bugfixes, this many changes might introduce some regressions, especially on non-developer devices.


samuelclay:
Holy moly, this is really something. The transitions are phenomenal. I'm going to release this to the beta channel immediately.

PR #1102, opened 2018-04-30
Android: Beta Fixes, Android 8 Compat, UI Affordances

dosiecki:
* Several fixes for reported crashes and bugs in the current Beta, mostly surrounding the new story list.
* Fixes and reversions for the 27.1.0 -> 27.1.1 support library update.  TL;DR: Google introduced an unprecedented number of servere bugs in 27.1.0, which we upgraded to in the last Beta and then worked around those changes as best we could. They gave in and fixed them in 27.1.1, requiring us to selectively revert some logic to the way it used to be.  (see https://developer.android.com/topic/libraries/support-library/revisions#27-1-1)
* Partial fixes for Android 8 compatibility. Android 8 made some essentially-undocumented changes to the lifecycle of background services and the freeze/thaw cycle for Activities.  This patch includes a few fixes for the known changes, and some logging to try and help us track down the rest. (not only does testing this require several hours of wating or intentional stressing of devices, I don't as yet have an Android 8 daily driver, so we are relying on user feedback)  Basic use still looks fine except for some edge cases, and no regressions spotted for pre-8 devices.
* A suite of quick-fixes to some places where we were relying on the exclusive use of colour to conveny intel information.  I ran the app through a full colourblindness simulator and confirmed where it was impossible to view/set intel training.  Changed some colours to increase contrast in the simulated diminished colour space.  Added a tweaked version of the design guidelines' background affordances for selected toggles.  Inserted icons into intel state chips.  Confirmed that all intel state can now be viewed from the intel dialogue, and all tag state can be viewed directly.  Will need to do a touch more work to present title and author state.
* Some lead-up work and refactoring to prepare for when we can tackle the bugs around the story Pager losing state.

Ready for immediate push to Beta.  

Contains DB changes - please ensure manifest version is bumped before building.


samuelclay:
I'm seeing debug statements like "27MB used, 4GB free".

samuelclay:
Feed icons are doing this weird caching thing where they show the wrong icon on scroll and then refresh to the correct icon. I think if there is no icon yet, don't show the old icon cached in the row cache.

dosiecki:
>  seeing debug statements like "27MB used, 4GB free".

Happens any time a debug flag is set.  Can always be disabled in `src/com/newsblur/util/AppConstants.java`

> where they show the wrong icon on scroll and then refresh to the correct icon. I think if there is no icon yet, don't show the old icon cached in the row cache.

I think this is a bug on launching the app on a device where a previous version was last launched a very long time ago using the old cache framework.  Added what I think should be a solid fix (but somewhat tricky to catch since it takes a few weeks to repro and I only had one device that stale).

Both fixed in #1103.

PR #1121, opened 2018-08-06
Android: Features and Compatibility

dosiecki:
Finally pushing a big set of updates now that the latest release is out and happy.

* Total rewrite of the platform ViewPager adapter to address the very long-standing issues around state loss when context switching while reading stories.  After much research, it appears that we weren't the only ones struggling to use the buggy platform impl, so went off and totally replaced it.  The new one understands the nature of story rivers being a dynamic list of items rather than an fixed-indexed dataset and correctly implements state save/restore for story fragments.  In most cases, things like scroll state are now preserved when rotating or context flipping.  This also fixes the root cause of the off-by-N problems we have had for years w.r.t. actions being performed on stories but not hitting the correct story or not being immediate reflected in the UI.
* Improved messaging and display behaviour for stories/feesd that don't support original text mode.
* Inserted some addtional logging to try and help debug the buggy WebView behaviour that GOOG rolled out some time over the last few months.
* New feature - feeds can now be renamed from the app.
* Bumped the target SDK version to 26.  This wasn't to gain features, but because in a few months GOOG will begin banning apps on older targets from the Play Store as a forcing function to push people away from older platform components.  For that reason, we also had to move our sync backend to a much newer (introduced in API 21) framework that works quite differently than the original one since API 1.  This new way of launching the syncs both in the background and when joined into the foreground should be compatible with the older platform versions while simultaniously not running afoul of the new background killer introduced in platform 28.
* Many other bugfixes and performance tweaks.

Two big components of this code are brand new and deal with some of the worst differences between platform versions.  This will need a good deal of testing on every live Android version from 5.0 - 8.1.  Though I have tested most of this on my 7.1 daily driver, the other versions need thorough soaking.   I would suggest we test this a bit more internally before even going to Beta.


samuelclay:
Wow, this is really something. With the transitions into the story list, this feels like a whole new app. I haven't had any issues, so I'm going to push this out as v9 beta 1.

samuelclay:
Deployed

PR #1123, opened 2018-09-05
Share to HTTPS

da2x:
Update some sharing endpoints.


samuelclay:
Thanks for doing this.

PR #1132, opened 2018-10-07
Docker:  docker-compose: Add front/back networks and persistent config storage

jeffballard:
This PR does two things for using NewsBlur with docker-compose:

1. Importantly, this PR puts Redis (along with postgres, elasticsearch and mongo) on an internal-only network. This is especially important as by default Redis isn't using a password, and if you happen to have this on a public IP address, miscreants from around the world will mess with your Redis instance.

2. This PR also puts as volumes the remaining configs for the different backends.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,6 +13,9 @@ services:
       - redis
     ports:
       - '8000:8000'
+    networks:
+      - front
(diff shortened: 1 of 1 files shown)


samuelclay:
Nice work

PR #1140, opened 2018-10-26
Attempt to fix fetching tumblr rss feeds.

jbouvier:
After some digging I was able to find a workaround for broken tumblr rss feeds. I've verified that using this user agent does in fact work, at least for the time being.

### DIFF
--- a/apps/rss_feeds/models.py
+++ b/apps/rss_feeds/models.py
@@ -11,6 +11,7 @@ import hashlib
 import redis
 import pymongo
 import HTMLParser
+import urlparse
 from collections import defaultdict
 from operator import itemgetter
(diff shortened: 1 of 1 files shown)


apps/rss_feeds/models.py
```diff
@@ -1086,6 +1087,12 @@ def calculate_scores(cls, facet):
     
     @property
     def user_agent(self):
+        feed_parts = urlparse.urlparse(self.feed_address)
```
jbouvier: It would probably be fine to skip `urlparse` and just do a `.find` directly on `self.feed_address`, but I was trying to be as specific as possible.

Issue #1144, opened 2018-11-04
Docker Compose Redis exits immediately


SiqingYu:
Here is the output of `docker-compose up`:
```
user@user-XPS-13-9350:~/repos/NewsBlur$ docker-compose up
Starting newsblur_mongo_1         ... done
Starting newsblur_elasticsearch_1 ... done
Starting newsblur_postgres_1      ... done
Recreating newsblur_redis_1       ... done
Recreating newsblur_newsblur_1    ... done
Attaching to newsblur_postgres_1, newsblur_mongo_1, newsblur_elasticsearch_1, newsblur_redis_1, newsblur_newsblur_1
postgres_1       | LOG:  database system was shut down at 2018-11-04 15:39:47 UTC
postgres_1       | LOG:  MultiXact member wraparound protections are now enabled
postgres_1       | LOG:  database system is ready to accept connections
postgres_1       | LOG:  autovacuum launcher started
mongo_1          | 2018-11-04T15:43:14.778+0000 I CONTROL  [initandlisten] MongoDB starting : pid=1 port=27017 dbpath=/data/db 64-bit host=5ff0ca4d1710
mongo_1          | 2018-11-04T15:43:14.778+0000 I CONTROL  [initandlisten] db version v3.2.21
mongo_1          | 2018-11-04T15:43:14.778+0000 I CONTROL  [initandlisten] git version: 1ab1010737145ba3761318508ff65ba74dfe8155
mongo_1          | 2018-11-04T15:43:14.778+0000 I CONTROL  [initandlisten] OpenSSL version: OpenSSL 1.0.1t  3 May 2016
mongo_1          | 2018-11-04T15:43:14.778+0000 I CONTROL  [initandlisten] allocator: tcmalloc
mongo_1          | 2018-11-04T15:43:14.778+0000 I CONTROL  [initandlisten] modules: none
mongo_1          | 2018-11-04T15:43:14.778+0000 I CONTROL  [initandlisten] build environment:
mongo_1          | 2018-11-04T15:43:14.778+0000 I CONTROL  [initandlisten]     distmod: debian81
mongo_1          | 2018-11-04T15:43:14.778+0000 I CONTROL  [initandlisten]     distarch: x86_64
mongo_1          | 2018-11-04T15:43:14.778+0000 I CONTROL  [initandlisten]     target_arch: x86_64
[...]

PR #1159, opened 2019-01-08
Fix: build postgres container without error

leophys:
Hi @samuelclay 

this PR addresses the issue I refer to in [this](https://github.com/samuelclay/NewsBlur/issues/1144#issuecomment-450338221) comment.

The db dump that is included has been changed as described in the aforementioned comment.


samuelclay:
Thank you!

PR #1190, opened 2019-05-28
ignore Chromium's dark mode settings, including html tag

caleb-allen:
chrome's dark mode updated to ignore the body tag, adding the html tag fixes it


samuelclay:
LGTM, deployed to beta channel as 9.0b2. If all goes well I'll push to prod.

caleb-allen:
Awesome, I'll keep an eye on logs

Issue #1143, opened 2018-11-04
Android: Preference for size of image thumbnails in story titles


samuelclay:
Same as #1142, but for Android.

After seeing Brett Simmons' new UI and thinking about what he got right, I think that the image previews next to story titles should have a third setting (apart from on/off) which is "Small thumbnail", which insets a small preview instead of taking up top to bottom space (and thereby shortening the title excessively, esp. on an iPad).

<img width="381" alt="timeline_light" src="https://user-images.githubusercontent.com/44229/47966213-f726f900-e01d-11e8-82ba-6e5836e50c65.png">

PR #1191, opened 2019-05-28
add image thumbnail size preference

caleb-allen:
addresses #1143


caleb-allen:
todo: need to confirm backwards compatibility moving from previous boolean thumbnail setting

samuelclay:
Whoa I never noticed this. Thanks!

PR #1193, opened 2019-05-31
Disable stories loading animation (if preference), to avoid memory leak

daanzu:
I've had trouble with Newsblur memory leaks for a long time, and recently Newsblur has been crashing its tab in Chrome multiple times a day. I tracked it down to this code. I'm not sure of the ultimate cause, but this seems to fix it for me (I have the preference set to false). I hope you can merge this and put it on the live site.


samuelclay:
Happy to merge this but I don't know why it would be this code. Just from looking at it it should be safe, although I can't tell for sure. But I see memory management issues with just leaving the web app open for a week, even if I don't load any feeds.

samuelclay:
Note that this is only for feeds and folders and not saved stories, shared stories, or searches.

samuelclay:
Deployed!

daanzu:
Great! Thanks for the quick response! I've had no crashes since the PR was merged.

PR #1205, opened 2019-06-18
remove clearing of fullscreen on window focus change

### LINKED ISSUE #1200: Android: Full screen mode only engages for 1 second on Pixel 2
Full screen mode turns on but then doesn't stay open for longer than a second.

caleb-allen:
android docs indicate that any system window can take focus from the activity, so we should not clear full screen in that case; the system will handle it if necessary (e.g. user leaves application, swipes navigation back up)

fixes #1200


samuelclay:
Nice, this worked well. Thank you!

PR #1229, opened 2019-08-20
Set folder name as tag when saving a story

### LINKED ISSUE #1168: Android: Record folder names as tags when saving a story
We do this both on the web and ios. Would be good to have on Android.

caleb-allen:
fixes #1168


samuelclay:
👍

Issue #1222, opened 2019-08-03
Android: Image feeds not loading


caleb-allen:
Possibly related to #1126 

Multiple user reports of image-only feeds failing to load:

> I'm having trouble with image only feeds. I'm getting "Sorry, the story's
text couldn't be extracted" in text mode and just broken picture with story
mode

> when I read a feed of the
xkcd comic, the comic does not come up no matter what I do. I always have
to view the original page.

samuelclay:
Yeah this is a long standing issue and one that I've experienced myself as well.

samuelclay:
Another recent thread:

> Still an issue today. In a normal browser it all shows properly so I really suspect something in the android app. Sometimes the switch to story / text does work, but not on all entries in the blog.
>
> The well-ordering principle direct link entry shows fine after toggling text/story, but switching to landscape or portrait removes the image again. Reproducably.
>
> Does that help you reproduce it too?
>
> My android is v9, on a Nokia 7 plus, patched to p1 from August the 1st.

samuelclay:
I still don't see XKCD images

caleb-allen:
Hm. But when you switch back to the text view it works?

Could you check out #1246 to see if a bigger margin fixes the issue on that first load?

If that doesn't help then the issue you're seeing probably isn't affected by the css padding.

caleb-allen:
Also, if it doesn't work, could you give me the webview version you're on?

Should be `WebView implementation` in developer settings

giganteous:
Hey! Didn't know this existed. I reported this on the forum, thanks for forwarding it here. 

My phone runs Android9, Webview version Chrome 77.0.3865.92, I can't opt for the "Android system webview" it is disabled for my user.

I have 2 image feeds that have this problem:
* xkcd (the feed you find when searching for it), _some_ articles never show, some after switching.
* http://www.karindaan.nl/beeldblog/feed/, switching doesn't help at all.

Both "worked in the past", I really do know that's not exact science. The web newsblur works fine. I sadly don't own another phone to test this.

Hope this helps. If you need more assistance: I'd be delighted.

samuelclay:
Hey @giganteous, I believe this is now fixed. If you want to go above and beyond, download and install Android Studio and then download this repo. If you open the [Android repo](https://github.com/samuelclay/NewsBlur/tree/master/clients/android) you'll be able to run the app directly on your device. Fun way to change settings to whatever you want, as well.

PR #1244, opened 2019-09-14
modify css to not overflow images on chromium webview

caleb-allen:
I think what was happening is that when a page loaded with a lot of content, the css styling could use sibling items to place images.

When an image alone was loaded, its width was being pushed outside the bounds of the parent by a single pixel, and the overflow style was to hide it.

This shifts the image to fit within the parent bounds.


caleb-allen:
This should fix #1222

PR #1246, opened 2019-09-17
image css modification test

caleb-allen:
continuing testing for #1222


samuelclay:
Seems to work with 4px.

PR #1259, opened 2019-10-23
fab: make sure a 'local' role exists by default.

Flameeyes:
This matches the instructions in README.md, and makes it possible for
fab to act on the local host without trying to connect to a remote
one.

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -60,6 +60,7 @@ except:
         'app'   : ['app01.newsblur.com'],
         'db'    : ['db01.newsblur.com'],
         'task'  : ['task01.newsblur.com'],
+        'local' : ['localhost'],
     }
 
 def do_roledefs(split=False):
--- a/fabfile.py
+++ b/fabfile.py
@@ -12,10 +12,11 @@ from vendor import yaml
 from pprint import pprint
 from collections import defaultdict
 from contextlib import contextmanager as _contextmanager
+import getpass
 import os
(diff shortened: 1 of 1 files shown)


fabfile.py
```diff
@@ -176,7 +178,8 @@ def ec2task():
 
 def ec2():
     env.user = 'ubuntu'
-    env.key_filename = ['/Users/sclay/.ec2/sclay.pem']
+    env.key_filename = '/Users/%(user)s/.ec2/$(user).pem' % {
+        'user': getpass.getuser()}
```
samuelclay: Since env.user is being overwritten with ubuntu, you'll have to use a variable that you set above.

fabfile.py
```diff
@@ -387,19 +390,19 @@ def change_shell():
         run('git clone git://github.com/robbyrussell/oh-my-zsh.git ~/.oh-my-zsh')
     sudo('chsh %s -s /bin/zsh' % env.user)
 
-def setup_user():
+def setup_user(username=getpass.getuser()):
```
samuelclay: I definitely don't want a function parameter initialized with some function. Just use the global variable env.user.

samuelclay:
Ok, we're moving away from fabric for setting up the repo. It'll soon be entirely Docker based and the developer working on it should have it done in the next month or two.

samuelclay:
See #1316

PR #1283, opened 2020-02-14
Add Edge as a browser option

jimmgarrido:
This adds the option to use Microsoft Edge as the default browser in the iOS app, and is based on the existing logic for using the Opera browser.


samuelclay:
Whoops, sorry, didn't notice this until now. Thanks for submitting this!

PR #1310, opened 2020-04-25
Android widget

sictiru:
Android Home Screen Widget for NewsBlur


sictiru:
Currently work in progress. The UI of the widget is 95% done and I'm still working on the background service feed set sync.

samuelclay:
Looking good! Let's cap it at 5 stories so it's less work on the server and less work for the phone. 

I'm getting an "Empty Feed" error after a restart and can't get it working again. When I launch the app, it freezes. Running it under the debugger gets it working again, but a restart then kills it again. Not sure if this is new.

samuelclay:
Looking good! Note that when you tap on a story, it should load in the "All Site Stories" river_stories reading-by-folder view, as that's where it's pulling from. And the Widget dialog should show feeds within their folders, allowing a user to select an entire folder. If this is a ton of work, let's ignore the widget dialog and just default it to "All Site Stories".

sictiru:
@samuelclay I made the changes you requested and the widget is ready for another review.

samuelclay:
Amazing, really enjoying this widget! Here's a few notes:

* Can we have the date field match the story list? It says "7 minutes ago" instead of "12:58pm"
* It's showing hidden stories. Somewhere there should be a calculate story score function, which basically gives a flag for hidden/unread/focus (-1/0/+1). [Here's for iOS](https://github.com/samuelclay/NewsBlur/blob/master/clients/ios/Classes/StoriesCollection.m#L536-L546) and [here's for web](https://github.com/samuelclay/NewsBlur/blob/master/media/js/newsblur/reader/reader_utils.js#L14-L30) and [here's for backend](https://github.com/samuelclay/NewsBlur/blob/master/apps/reader/models.py#L866-L875). This functionality exists on Android, I just don't know where. But the "All Site Stories" folder view doesn't show hidden stories and the widget shouldn't either.

samuelclay:
Alright neat! I won't be able to look at this until next Monday as I'm traveling for the week (not by choice) and I left my Pixel 2 at home. I suppose even if there are small bugs, at this point it's looking good and is ready to ship. I'm going to release this to the beta channel.

Splike:
It looks good, but clicking a on a story will often open the "All Stories" and show "Loading...", which doesn't go away.

When I do succeed in opening a story through the widget, I can go to my homescreen to click another story via the widget, but the first story will be shown instead of the new one.

Also, it would be nice if we can choose a specific folder for the widget, instead of only "All stories".

samuelclay:
I agree it would be nice if the widget's feed chooser dialog was in the same order and layout as the feed list, in that it shows both folders and feeds inside folders.

sictiru:
I might be misunderstanding this but [Widget Feed List](https://ibb.co/VHm7Sgj) already behaves like [Main Feed List](https://ibb.co/86qnKqx) and shows both folder and feeds inside folders?

Will look into the other two issues.

samuelclay:
Ohh you're right, I was looking at an older build.

Splike:
Ah! I just added the widget to my homescreen and didn't know I had to go to the widget menu to change the displayed feeds.

I do see some improvements for the user experience:
- Clicking the Newsblur logo at the top of the widget should open the app.
- The top right corner of the widget should have a cogwheel icon, clicking it takes you to the widget organisation screen.
- Folders in the organisation screen should be collapsible and also be able to be check on/off like the different feeds do.
- I'd also like to have an option to make the background of the widget completely transparent, as that's how I display all my widgets.

PR #1320, opened 2020-06-07
Android widget 2

sictiru:


samuelclay:
Released as v10.0b2.

samuelclay:
Having an issue when I tap on a story and it isn't found (so I just see loading... forever). Seems to happen when I was recently reading a story, exit out to the home screen, and then spot a new story that came in.

samuelclay:
When I hit the back arrow, the story list shows the original stories from the first load, not the newly-loaded stories. So it seems it's serving stale data.

Splike:
Didn't notice the new thread about the widget. I will repost my response from #1310 here:

>Ah! I just added the widget to my homescreen and didn't know I had to go to the widget menu to change the displayed feeds.
>
>I do see some improvements for the user experience:
>
>- Clicking the Newsblur logo at the top of the widget should open the app.
>- The top right corner of the widget should have a cogwheel icon, clicking it takes you to the widget organisation screen.
>- Folders in the organisation screen should be collapsible and also be able to be check on/off like the different feeds do.
>- I'd also like to have an option to make the background of the widget completely transparent, as that's how I display all my widgets.

samuelclay:
> > * Clicking the Newsblur logo at the top of the widget should open the app.

Agreed

> > * The top right corner of the widget should have a cogwheel icon, clicking it takes you to the widget organisation screen.

I'd rather not have that cog showing. If you want to configure it, hit the logo and then open the widget preferences. The intention is that you're not in there more than once in a while.

> > * Folders in the organisation screen should be collapsible and also be able to be check on/off like the different feeds do.

I don't think we need collapsible folders in the widget dialog. It's similar to the Organize and Mute Feeds dialog in that it's an overview, not an everyday dialog. However, tapping folder titles does indeed turn feeds on and off.

> > * I'd also like to have an option to make the background of the widget completely transparent, as that's how I display all my widgets.

Agreed!

PR #1327, opened 2020-06-14
Android widget 3

sictiru:
Fixing [loading issue](https://github.com/samuelclay/NewsBlur/pull/1320#issuecomment-640791190) and adding [widget transparent background](https://github.com/samuelclay/NewsBlur/pull/1320#issuecomment-641353172) option.


samuelclay:
Can we change the "Tap to setup in NewsBlur" to say "Loading..." when it's just loading? I see that message every time and I know it's already setup. It should say it when necessary though.

samuelclay:
By the way, I think we should switch to a system like what @Dejal and I are using, and that's to have a ticket for the overall feature (which I've just made here: ) and you'll just commit to a branch. I'll merge in the branch on my own so it's a bit less work for you, and that we have have a single ticket.

Feel free to create feature branches as necessary, but please make a sictiru branch as your main.

PR #1337, opened 2020-06-29
Android api 28 update

sictiru:


samuelclay:
You can just merge this directly into the sictiru branch without a PR. I'll update the ticket if I find anything amiss.

samuelclay:
What version fo Gradle are we using? I tried to upgrade but got gradle errors.

samuelclay:
Should we be on Gradle 6.0?

sictiru:
@samuelclay Currently we're using Gradle 5.6.4. I tried updating to the latest version but got an plugin error and will need to familiarize myself today with it before removing it. Also after this is sorted out we'll be on the latest 4.0 Android Gradle Plugin.

sictiru:
Android Gradle 4.0 comes with a different annotation processor and since the project uses a very old version of ButterKnife for view injection I would need to update ButterKnife which requires some refactoring. 
Because ButterKnife is deprecated in favor of the new view binding how would you like to proceed? I could upgrade ButterKnife or spend more time removing and replacing it with view binding.

samuelclay:
You're better equipped to make that call. I'll say this, I'm knee deep in updating old dependencies on the backend right now, so if there's technical debt that needs to be paid off, I'm in that mindset.

Issue #1308, opened 2020-04-24
signup not possible - recaptcha wrong after new installation


xanthos84:
How can I enter the correct recaptcha API keys? settings.py ? and how?
Thanks

JasonGhent:
Form submit from `GET /account/signup` page was not respecting the `ENFORCE_SIGNUP_CAPTCHA` flag for me. Signing up from the signup widget at `GET /` worked for me on latest using docker-compose and the default local_settings.py.template.

samuelclay:
This will be addressed in #1316, but yeah you just need to signup on the homepage.

PR #1360, opened 2020-09-01
Respect ENFORCE_SIGNUP_CAPTCHA in profile signup

JasonGhent:
Tested with Cypress [here](https://github.com/JasonGhent/NewsBlur/blob/cypress/Makefile), which will produce a video output:

index_spec js


JasonGhent:
Linking to #1308.

samuelclay:
Nice work, deployed.

samuelclay:
I should note, since you're using Docker, we have a full-fledged and vetted docker install being worked on in the jmath branch. Check it out!

PR #1395, opened 2020-12-22
apps/reader/views: Send last_read_date when retrieving read stories

alistair23:
When retrieving the list of read stories from the server return the
last_read_date. This is useful for clients that re-downloaded stories
when performing a sync.

Unfortunately I am unable to figure out how to test this.

Signed-off-by: Alistair Francis <alistair@alistair23.me>

### DIFF
--- a/apps/reader/views.py
+++ b/apps/reader/views.py
@@ -1260,7 +1260,8 @@ def load_read_stories(request):
     
     story_hashes   = [story['story_hash'] for story in stories]
     story_feed_ids = list(set(s['story_feed_id'] for s in stories))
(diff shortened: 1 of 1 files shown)


alistair23:
Ping!

samuelclay:
Whoa, sorry about the wait on this, I somehow missed this when you submitted it. Anyway, great idea but I'm not sure what you're looking to do. You add a `last_read_date` to an individual story from the read_stories endpoint, but you already know its read date as its a read story. And feeds already come with a last_read_date in the `/reader/feeds` endpoint. I accidentally merged this but I'll have to revert until I can figure out what exactly we want to add.

PR #1402, opened 2021-01-24
Makefile: specify SHELL bash

starsep:
Without it I get an error:
/bin/sh: 1: [[: not found
Which is ignored due to using leading dash `-` at https://github.com/samuelclay/NewsBlur/blob/3ad5295a33bb6fd469f8a64f0da6094656ad2da1/Makefile#L16

The same issue exists also in docker_django3.0


samuelclay:
Thanks, this is now downstream in the `dashboard3` branch.

PR #1428, opened 2021-03-05
Move WebSub callbacks to HTTPS

da2x:
The HTTP endpoint just redirects to HTTPS anyway, so let’s skip a redirect (clients may not even implement it?) and protect the integrity of the payload.


samuelclay:
Nice, thanks for doing that. How did you catch this?

da2x:
I made stupid mistake in a hub implementation, and while debugging it I noticed that Newsblur and Feeldy were alone in using HTTP in its callbacks. Everyone else seems to have migrated to HTTPS.

PR #1448, opened 2021-04-01
Add XML declaration to OPML output (sets utf-8)

da2x:


samuelclay:
Looks good, thanks for adding this.

PR #1541, opened 2021-08-27
fix YouTube video duration calculation

louis-pre:
The current implementation does not play well with longer durations than 60 seconds because the results of the divisions to obtain hours and minutes aren't floored before being multiplied.

Example:
- 3679 seconds is converted to '1.0219444444444445:00:00' instead of '1:01:19'
- I believe this is linked to [this forum topic](https://forum.newsblur.com/t/youtube-video-length/8860)


louis-pre:
Is there anything I should do to improve this PR?

louis-pre:
Closing because this is polluting my backlog

samuelclay:
Hey there, sorry for the wait. I was on paternity leave since July and now I'm back to work. This PR looks terrific! I'll merge and deploy right now, you should see results come in for future stories as well as some backdated stories (over the next few hours).

PR #1547, opened 2021-09-16
Metrics

jmath1:


Makefile
```diff
@@ -1,6 +1,7 @@
 SHELL := /bin/bash
 CURRENT_UID := $(shell id -u)
 CURRENT_GID := $(shell id -g)
+newsblur := $(shell docker ps -qf "name=newsblur_web")
```
samuelclay, line 4: Rename to a description variable. A `newsblur` variable in the Makefile isn't as description as newsblur_docker_pid or whatever this command does.

samuelclay: I see its just moved from below, but let's rename anyway

ansible/playbooks/setup_mongo.yml
```diff
@@ -19,4 +19,5 @@
     - {role: 'node-exporter', tags: ['node-exporter', 'metrics']}
     - {role: 'mongo-exporter', tags: ['mongo-exporter', 'metrics']}
     - {role: 'monitor', tags: 'monitor'}
+    - {role: 'flask_metrics', tags: 'flask_metrics'}
```
samuelclay, line 22: Use hyphen instead of underscore to match style above

config/fixtures/bootstrap.json
```diff
@@ -15,14 +15,6 @@
...
         }
     },
-    {
-        "pk": 3,
-        "model": "sites.site",
-        "fields": {
-            "domain": "nb.local.com",
-            "name": "NewsBlur"
-        }
-    },
```
samuelclay, line 25: Why remove this? I use it locally as I still like to use the old nb.local.com.

newsblur_web/docker_local_settings.py
```diff
@@ -170,8 +170,8 @@
 SERVER_NAME = "nblocalhost"
 NEWSBLUR_URL = os.getenv("NEWSBLUR_URL", "https://localhost")
 
-if NEWSBLUR_URL == 'https://nb.local.com':
-    SESSION_COOKIE_DOMAIN = ".nb.local.com"
+if NEWSBLUR_URL == 'https://localhost':
+    SESSION_COOKIE_DOMAIN = "localhost"
```
samuelclay, line 174: Ahh this is why nb.local.com was removed above. I think we can keep it in while preserving this change.

package-lock.json
```diff
@@ -0,0 +1,3 @@
+{
+  "lockfileVersion": 1
```
samuelclay, line 2: Is this file top level? Probably shouldn't be here.

samuelclay (commented):
Nice work, just a few changes before we ship it.

PR #1548, opened 2021-09-19
fix(docker): Remove duplicate YouTube API key setting

louis-pre:
## Context
The Youtube API key setting appears twice in the settings file for docker (lines 74 and 180)

## Changes
Remove the occurrence that seems out of place


samuelclay:
Thanks!

PR #1583, opened 2021-12-22
Update docker-compose.yml

Linecutterx:
7.16.2 has the update(s) needed to avoid log4j exploitation


samuelclay:
Good catch, thanks!

PR #1749, opened 2022-11-08
Add Docker-compose-plugin to Prerequisites.

xenspidey:
Added the missing docker compose plugin when not using the docker upstream repository. 

Touched up the markdown formatting a little

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,16 +46,19 @@
    [See the complete list](https://github.com/samuelclay/NewsBlur/tree/master/media/js).
 
 
-### Prerequisites
-    * Docker
-    * Docker-compose
+## Prerequisites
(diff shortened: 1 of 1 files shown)


danielhoherd:
I ran into this bug too, but given that the compose plugin is only supported on linux, and given that docker-compose is already a prerequisite, is there any benefit to using `docker compose` vs `docker-compose` ?

samuelclay:
I like the formatting changes but the build context change and Dockerfile change don't work for me. Also the top change is the linux-only docker compose plugin, which I believe is part of the Docker Compose install. I don't think we need these changes but if you want to rewrite them so they are a section specific to Linux, I would be glad to reconsider!

PR #1772, opened 2022-12-24
fixes issues with subdomain hosting of user blurblogs

usmcamp0811:
This is a super simple fix for the problem I outlined in [this forum post](https://forum.newsblur.com/t/what-is-the-container-endpoint-middleware-used-to-connec-to-the-public-blurblog/9783). Instead of looking at the current host and checking to see if there is exactly 2 periods.. I am saying if there are 2 or more then just return the first split is presumed to be the sub (maybe sub sub) domain that we care about. 

It works for me with the NewsBlur being served at `news.mydomain.com`, but I could imagine this working with even more nested sub domains.. ¯\_(ツ)_/¯

### DIFF
--- a/apps/reader/views.py
+++ b/apps/reader/views.py
@@ -86,7 +86,7 @@ ALLOWED_SUBDOMAINS = [
 
 def get_subdomain(request):
     host = request.META.get('HTTP_HOST')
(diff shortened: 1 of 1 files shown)


samuelclay:
Nice, glad this is here!

PR #1778, opened 2023-01-25
Multiple changes to enable running on AlmaLinux 8 with podman and some Youtube RSS fixes

softmanro:


samuelclay:
The only change in this PR I would like to introduce is the subdomain change, which I'll add now. But the rest of this is better suited for your own repo forked from this one, as it's too niche to warrant the change for all developers. I think the podman references that will be applied to non-podman deployments could be fixed. And I'd pass `change_ua_youtube` directly instead of the nonlocal scope. But I'm not comfortable including a change to turn off the NewsBlur user agent.

Thanks for providing this but I think it's better for you to keep most of this on your own fork. I'll support you if I can to make your fork successful.

I made the subdomain change in 36beb2fb4.

Issue #1397, opened 2021-01-03
Screenshots for F-Droid


mondlicht-und-sterne:
You can add Screenshots for F-Droid, too, if you want.

https://f-droid.org/en/docs/All_About_Descriptions_Graphics_and_Screenshots/

https://gitlab.com/snippets/1895688

samuelclay:
We don't maintain the F-Droid repo, but if there's somebody who does, screenshots are available in 

https://github.com/samuelclay/NewsBlur/tree/master/media/img/android

shuvashish76:
@IzzySoft request to provide fastlane metadata from PlayStore & merge with F-Droid.

PR #1799, opened 2023-05-24
Add fastlane metadata for F-Droid

shuvashish76:
Addresses #1397


samuelclay:
This seems great, I'm just going to move it into the clients/android/fastlane directory

shuvashish76:
>  move it into the clients/android/fastlane directory

But as per the F-Droid doc, it has to be in root directory.
@linsui Request to check if its okay to use _clients/android/fastlane_ directory for F-Droid.

linsui:
It must be in the root directory currently.

shuvashish76:
@samuelclay `v13.0.0` available on F-Droid but still we're missing metadata because they're in the wrong place 😔

Issue #1793, opened 2023-04-16
Backend: Require authorization for viewing feeds with few subscribers


aladh:
## Problem

All feeds are currently publicly accessible via their numeric feed IDs, making it easy to enumerate and access feeds added by other users. This situation violates users' expectations of privacy and security. Access to some RSS feeds should be limited to the user who added them, as they are meant to be private.

This issue has been raised in previous forum discussions:

- https://forum.newsblur.com/t/feeds-are-not-private-anyone-can-see-any-feed/4579
- https://forum.newsblur.com/t/private-feeds-are-public-for-everyone-to-see/2490

For information on potential exploitation, see https://owasp.org/www-community/attacks/Forced_browsing.

## Proposal

Treat feeds with fewer than N subscribers as private, requiring authorization for access. Limit access to existing subscribers, similar to email newsletters. If the number of subscribers exceeds N, the feed can revert to the current public behaviour. The value of N could align with the threshold for a feed's appearance in search results.

Here's a small diff to illustrate the proposed solution:

```diff
diff --git a/apps/reader/views.py b/apps/reader/views.py
index 557c9e27d..515d01d76 100644
--- a/apps/reader/views.py
+++ b/apps/reader/views.py
@@ -671,6 +671,10 @@ def load_single_feed(request, feed_id):
     if feed.is_newsletter and not usersub:
         # User must be subscribed to a newsletter in order to read it
         raise Http404
+
+    if feed.num_subscribers < 10 and not usersub:
+        # This feed might be private so user must be subscribed in order to read it
+        raise Http404
     
     if page > 400:
         logging.user(request, "~BR~FK~SBOver page 400 on single feed: %s" % page)
```

aladh:
Hi @samuelclay! Can I get your thoughts on this? Thanks!

PR #1817, opened 2023-09-02
Prevent unauthorized access to feeds with a single subscriber

aladh:
Fixes https://github.com/samuelclay/NewsBlur/issues/1793


aladh:
@samuelclay would you mind taking a look at this?

samuelclay:
I'm afraid this might break something but I'm happy to give it a try. Eventually I'll have to build a way to preview feeds using the url, so if somebody has the url or it shows up in autocomplete, they can preview the feed.

PR #1871, opened 2024-06-26
Fix not forcing Android clients to use `refresh_feeds`

quad:
This patches a long outstanding bug in 50e88b2bafd8b183b364484b57d3ed455db7b27a.

### DIFF
--- a/apps/reader/views.py
+++ b/apps/reader/views.py
@@ -337,7 +337,7 @@ def load_feeds(request):
         return load_feeds_flat(request)
 
     platform = extract_user_agent(request)
-    if platform in ["iPhone", "iPad", "Androd"]:
+    if platform in ["iPhone", "iPad", "Android"]:
         # Remove this check once the iOS and Android updates go out which have update_counts=False
         # and then guarantee a refresh_feeds call
         update_counts = False


satmandu:
Thanks for this!

samuelclay:
@sictiru I agree with this ticket, we should do the right thing and call `/reader/feeds?flat=true&refresh_counts=false` and then immediately call `/reader/refresh_feeds/` to get the right count. On iOS, we show a "Counting is difficult..." message. I kinda remember making this call to get Android shipped a tiny bit faster, but it means Android is slow to load. Let's get this fixed!

samuelclay:
Oh, this PR isn't the way to do it. But we can keep this PR as the ticket for it for now. I'm not going to be merging this, since this is fixing it in the wrong spot, but it's helpful to have as a reference, so let's just use this as the Android ticket.

samuelclay:
Ohh, looking more carefully at this PR (and not the commit that I mistakenly thought was this PR's code), I realize that it is doing the right thing already. `Androd` is the 6 letter version of Android that i use in `extract_user_agent` since it makes the log a bit cleaner (iPhone is 6 letters, so they take the same size).

https://github.com/samuelclay/NewsBlur/blob/a5229ce5cbc19f7d84d68ad92ae0a26993f35ada/utils/user_functions.py#L139

samuelclay:
We can continue discussing in #1872, but this PR isn't correct because that's just the platform log.

PR #1917, opened 2024-11-17
chore(node): update nodejs stuff

viq:
Node image to LTS 22.11.0
Chase rename of mercury-parser -> parser
npm update and npm audit fix

With this my instance running in docker (that I started using today) is able to start the `newsblur_node` container, no longer complains on all articles about being unable to get text of the story, and the Text view of feed entries works.


samuelclay (approved):

samuelclay:
Yep, about time. Thanks for doing this!

PR #1952, opened 2025-05-13
Add Readwise to third-party sharing services

jakevictor:
This PR adds [Readwise Reader](https://readwise.io/read) to the available third-party sharing services.

The `https://readwise.io/save` URL was taken from the Readwise Reader bookmarklet code found [here](https://read.readwise.io/resources) (link may require being logged into Readwise Reader).

This is my first contribution to NewsBlur, so please let me know if I'm doing anything crazy.


samuelclay:
Amazing, this is great! One addition is something needs to call `send_story_to_readwise`. See how we call its neighbors and you'll spot the pattern.

jakevictor:
It looked like that was already handled dynamically based on the name of the third party:
https://github.com/samuelclay/NewsBlur/blob/e2ef0666ba122365816208a06fa252e31fac528b/media/js/newsblur/reader/reader.js#L2620-L2626

jakevictor:
@samuelclay, it looked like the calling of `send_story_to_readwise` was already handled dynamically based on the name of the third party:
https://github.com/samuelclay/NewsBlur/blob/e2ef0666ba122365816208a06fa252e31fac528b/media/js/newsblur/reader/reader.js#L2620-L2626

I also ran the branch locally, and it seemed to properly call `send_story_to_readwise` and add the URL to my Readwise library when I clicked the new icon.

PR #1975, opened 2025-10-07
Don't prepend `https://` to newsletters' anchor links.

nfvs:
Currently any anchor links like `#footnote` in newsletters' HTML gets `https://` prepended, leading to `https://#footnote`, which is wrong. This fixes it by keeping them as-is.

### DIFF
--- a/utils/scrubber/__init__.py
+++ b/utils/scrubber/__init__.py
@@ -297,7 +297,7 @@ class Scrubber(object):
 
     def _clean_path(self, node, attrname):
         url = node.get(attrname)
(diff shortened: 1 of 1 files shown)


samuelclay:
Good thinking, thanks for submitting this.

nfvs:
> Good thinking, thanks for submitting this.

Np, scratching my own itch. Thanks for the feature! 😊

May I ask when this might make it to production?

samuelclay:
Sorry for not responding earlier, but the answer was probably within the day.

PR #1986, opened 2025-10-22
Allow `id=` attributes in the scrubber.

nfvs:
Without it anchor links in newsletters won't work.


samuelclay:
I agree, this is a good idea. I'll deploy it soon with the upcoming Ask AI feature.

PR #1987, opened 2025-10-23
Modified FeedReading.java and SocialFeedReading.java

nbd-boss:
Hello，
These two database operations are executed on the UI thread, which may block the interface. To avoid potential performance issues, they should be moved to a background thread.


sictiru:
Ideally this would be done in a ViewModel but until then something like this would be an improvement
```
lifecycleScope.launch(Dispatchers.IO) {
    val feed = dbHelper.getFeed(fs.getSingleFeed())
    withContext(Dispatchers.Main) {
        if (feed != null) {
            UIUtils.setupToolbar(this@YourActivity, feed.faviconUrl, feed.title, iconLoader, false)
        } else {
            finish()
        }
    }
}
```

You can convert these classes to Kotlin. 
We're in the process of moving quite a few things to coroutines and off the main thread.

sictiru:
Code has been updated to Kotlin and the queries moved off the main thread

PR #2001, opened 2025-11-27
Android: Add support for HTTPS & HTTP for self hosted server 

### LINKED ISSUE #1924: Android: Login is broken when using self-signed certificates
The root CA has been imported in the system trust store and and the web version in chrome works file without any certificate errors, but the Android app fails with:

```
1736320032.476 14271 32677 18459 I o1.b    : API call failed, pausing before retry number 2
1736320035.519 14271 32677 18459 E CONSCRYPT: ------------------Untrusted chain: ----------------------
...
1736320035.531 14271 32677 18459 E o1.c    : Error (java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.) calling https://newsblur.local/api/login
1736320035.531 14271 32677 18459 E o1.c    : javax.net.ssl.SSLHandshakeException: java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.
1736320035.531 14271 32677 18459 E o1.c    : 	at com.android.org.conscrypt.ConscryptFileDescriptorSocket.startHandshake(ConscryptFileDescriptorSocket.java:239)
```

Android 10-14, latest NewsBlur version

acendrou:
This PR allows the use of a custom certificate for a self-hosted NewsBlur server with an HTTPS connection and enables the use of an HTTP connection to a NewsBlur server. (Fixes #1924)

### DIFF
--- a/clients/android/NewsBlur/app/src/main/res/xml/network_security_config.xml
+++ b/clients/android/NewsBlur/app/src/main/res/xml/network_security_config.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <network-security-config xmlns:tools="http://schemas.android.com/tools">
-    <base-config>
+    <base-config cleartextTrafficPermitted="true"
+        tools:ignore="InsecureBaseConfiguration">
         <trust-anchors>
             <!-- Trust preinstalled CAs -->
             <certificates src="system"/>


samuelclay:
You know, honestly, we need to figure out how to do HTTPS for self-hosted servers. My development environment is a self-hosted server, and I have to type `thisisunsafe` all the time to access it in Edge. The Chrome/Edge flags that used to allow localhost on HTTPS are gone now, so I'd love to learn what devs are doing about this. But I'm speaking about the larger case. 

Does this look good @sictiru?

Also, @claude, no need to review the code or write any code, but what do developers do to have https with self-signed certs to run NewsBlur in development (which is no different than self-hosting and figuring out tailscale/dns). Notice that we self-sign a cert (see `make cert`) but it's annoying and the mobile apps don't support it.

samuelclay:
Ok, from claude above my takeaway is that you should be using tailscale and I should switch to mkcert, which may work in your case but requires the installation of a root CA. I think that was the rec from #1924. I'm not sure if `acceptsUserCertificates` is a pre-req to accomplish that though.

acendrou:
Thanks for your response. I removed some useless code, which is why I force pushed to my branch. 

I generated the X.509 certificates using OpenSSL (although other software, such as mkcert, could have been used), as well as the following private keys:
- root CA
- intermediate CA
- certificate and private key corresponding to my domain name hosting newsblur.

I installed the NewsBlur server using the intermediate CA, certificate and private key.
I installed the root CA on my Android phone as a trusted CA for the user. This is a keystore that is separate from those managed by Android (system).
Currently, the NewsBlur Android app is configured by default to only access the CA managed by Android. My pull request simply allows the NewsBlur app to access the two Android keystores: user and system.
This way, it can trust my own root CA.

My NewsBlur server, along with other servers, is on a private LAN. I access it via the internet using a StrongSwan IPSEC VPN, which gives me control over both ends. Therefore, I do not use Tailscale.

sictiru:
The proposed changes weaken the overall app security. Since API 28+ all traffic on Android needs to be over HTTPS and exceptions can be added via XML but only for static domains. Trusting all of the user certs is also not recommended.
Both changes are flagged by the IDE and the code explicitly ignores them.

A solutions that I'm considering is using two `OkHttp` instances - default one that talks only to newsblur.com and another one that will handle system certs + imported certs for self hosted domains. Once it's implemented I'll post here the proposed PR for testing.

sictiru:
@acendrou https://github.com/sictiru/NewsBlur/pull/29

acendrou:
I've just tested it (https://github.com/sictiru/NewsBlur/pull/29) and it works! Many thanks

sictiru:
Great. It will probably go out in the next release.

sictiru:
https://github.com/sictiru/NewsBlur/pull/29

PR #2012, opened 2025-12-26
Fix Share Extension subfolder selection and add site/folder to notifications

samuelclay:
This adds an extra notification when a user adds a site (`Adding [site] to [folder]`) from the share extension widget. It also fixes the folder chooser so adding a feed to a folder actually adds it to the folder and not the top level.


samuelclay:
I've been using the ios 18 simulator to test, it's worked well.

PR #2022, opened 2026-01-07
Custom icons on iOS

samuelclay:
This PR only shows the custom icons, but it shows them everywhere they are displayed (feed list, story titles list, etc).

### DIFF
--- a/clients/ios/Classes/FeedsObjCViewController.m
+++ b/clients/ios/Classes/FeedsObjCViewController.m
@@ -374,7 +374,9 @@ static NSArray<NSString *> *NewsBlurTopSectionNames;
 
     NSUserDefaults *preferences = [NSUserDefaults standardUserDefaults];
 //    NSLog(@"Show feeds after being read (%@): %d / %@ -> %@", feedIdStr, [preferences boolForKey:@"show_feeds_after_being_read"], [self.stillVisibleFeeds objectForKey:feedIdStr], self.stillVisibleFeeds);
-    NSIndexPath *visiblePath = [self.stillVisibleFeeds objectForKey:feedIdStr];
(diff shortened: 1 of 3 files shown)


Dejal:
Is there an easy way to merge/pull this into my local repo working copy, so I can test it?

Also, if you're going to add functionality to the iOS/Mac app, I recommend getting Claude to use Swift. Live in the now!

samuelclay:
That's a fair point on Obj-C, but these are existing files and converting those is a huge task that we don't need to do yet. Maybe we rewrite each view piecemeal one day this year.

As for your local repo, you should be able to handle that by pulling this branch and merging it locally.

Dejal:
I don't mean rewriting the existing view; I agree that's a bad idea. (I have Swift wrappers for new code where appropriate). But when writing a new class (like `CustomIconRenderer` in this example), it should be done in Swift. I assume Claude can be instructed to write new classes in Swift but update existing ones as-is.

I took a look at the code, but there are several conflicts, including with `MainInterface.storyboard`, which is too difficult to merge. Can you please use the `dejal-alpha` branch for iOS changes. It has extensive changes from the `main` and `dejal` branches, and all new work should be done there.

samuelclay:
You bet, I'll make those changes and merge it in. You know I'm not doing either manually, the obj-c to swift rewrite will be trivial, and the dejal-alpha merge often requires no input. I'll sometimes ask Claude to use the `AskUserQuestion` tool to decide merge conflicts, but it can accumulate its own context better than I can.

samuelclay:
Ok, its now a clean PR on `dejal-alpha`
Read All the Stories

samuelclay:
Closing in lieu of #2024, which has a bunch more (and more coming) iOS 26 changes to be merged into `dejal-alpha`

PR #2057, opened 2026-02-12
Add story clustering to group duplicate stories across feeds

samuelclay:
## Summary
- New `apps/clustering` module that groups stories with matching or similar titles from different feeds using exact normalized title matching plus fuzzy Jaccard similarity on significant words
- Clusters stored in Redis (`sCL:` / `zCL:` keys, 14-day TTL) and displayed as always-expanded inline source list below the representative story in both river and single-feed views
- Celery task `ComputeStoryClusters` triggered after feed updates for feeds with premium subscribers, rate-limited to once per 6h per feed
- Briefing integration: shared `normalize_title()`, pre-computed cluster lookups in `_find_duplicate_stories()`, and cluster annotations in AI summary prompts
Screenshot 2026-02-11 at 9 50 56 PM

## Test plan
- [ ] Open All Site Stories in river view, verify cluster sources appear inline below stories that have duplicates across feeds
- [ ] Click into a single feed that has clustered stories, verify clusters also appear there
- [ ] Verify cluster quality: all grouped stories should be about the same event from different feeds
- [ ] Check dark theme styling of cluster source rows
- [ ] Verify non-premium users do not see clusters
- [ ] Confirm Celery task runs after feed updates (`docker logs newsblur_celery | grep Clustering`)

---
Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,213 @@
+# NewsBlur Architecture
+
+NewsBlur is a personal news reader that brings people together to talk about the world.
+It is a full-stack RSS reader with intelligence (train stories you like/dislike),
+social features (follow friends, share stories), and AI capabilities (ask AI, briefings).
+
+## High-Level Overview
+
+```
+┌──────────────┐    ┌───────────┐    ┌───────────────┐
+│   Browser    │───▶│  HAProxy  │───▶│  Nginx        │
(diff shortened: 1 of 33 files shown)


samuelclay:
Staff only but ready to roll and see how it does!

PR #2063, opened 2026-02-14
Chevrons animate in the direction they point

quad:


samuelclay:
Hah, amazing, thanks for being on it. I take it you've come across this before and it bothered you enough to fix?

quad:
> Hah, amazing, thanks for being on it. I take it you've come across this before and it bothered you enough to fix?

I'm not sure what you mean? I opened newsblur.com and there they were— three chevrons animating in the wrong direction! I inspected them in devtools and it was, as I expected, a trivial fix.

PR #2086, opened 2026-03-14
Drop the mutable default in canonical_string

tejasae-afk:
Replaces mutable default in canonical_string() from utils/S3.py to prevent cross-call state leaks.

Reason: Mutable defaults are shared across calls and usually turn into state leaks.

Validation: `utils/S3.py`.

Backward compatibility: This changes a public function signature by replacing mutable defaults with None guards. Call sites stay source-compatible, but callers introspecting defaults will see None now..

Context: py-mutable-default at utils/S3.py:32.

### DIFF
--- a/utils/S3.py
+++ b/utils/S3.py
@@ -29,7 +29,11 @@ AMAZON_HEADER_PREFIX = "x-amz-"
 
 
 # generates the aws canonical string for the given parameters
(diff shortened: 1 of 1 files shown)


samuelclay:
Oof, thanks for catching this!

